PR #7, opened 2019-03-18
Port "block_basic_core"

marcus8448:
Split up and ported "block_basic_core"
* Tin decoration block
* Tin wall block
* Copper ore
* Tin ore
* Aluminum ore
* Silicon ore
* Copper block
* Tin block
* Aluminum block
* Silicon block
(Nothing has textures yet.)

Note: Can I have write access? (So I don't have to make these PRs)


JoeZwet:
Given you write access.

PR #20, opened 2019-03-31
Fix a compile error with containers.

AlexIIL:
This could probably be done in a cleaner way.


JoeZwet:
thanks for the PR, I'll add you to the contributor list so u dont have to make a pr if u help out again.

PR #28, opened 2019-06-24
Update LibBlockAttributes to 0.4.5

AlexIIL:
A few notes:

- The refinery container + screen have errors compiling ATM - they don't seem to be registered in GalacticraftContainers or GalacticraftClient ? (Plus it references the OxygenCollectorContainer instead of the RefineryContainer - is that a typo?
- I've changed the way that containers + screens are created - hopefully that's ok?
- I've made most of the machines sync their status in the container via `Property`'s rather than tick on the client. (Theoretically this should prevent desyncs between multiple clients and the server)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -42,7 +42,7 @@ dependencies {
     modCompile("io.github.prospector.modmenu:ModMenu:${project.modmenu_version}") { transitive = false }
 //    modCompile "cotton:cotton:${project.cotton_version}"
     modCompile("io.github.cottonmc:cotton-energy:${project.energy_version}") { transitive = false }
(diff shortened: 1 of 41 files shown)


AlexIIL:
> The refinery container + screen have errors compiling ATM

Are you OK fixing this + else that isn't quite working properly yet?

JoeZwet:
yeah I'm going through and updating everything to 1.14.3 anyway so I'll just fix that then

AlexIIL:
Oh and just for future reference: LBA 0.4.x doesn't change `SimpleFixedItemInv.getInvStack` to return mutable ItemStacks, but you can always drop-replace it with `DirectFixedItemInv.get()` which works much closer to a vanilla inventory.

PR #37, opened 2020-06-13
Full 1.16 support

hYdos:
This PR ports all the features to the 20w20a features to 1.16-pre5.


hYdos:
just going to add some bug fixes here and there

PR #40, opened 2020-06-15
Launch Pads

hYdos:


JoeZwet:
are we able to get an easy way to interact with the launch pad from other machines or this that out of scope for this pr?

hYdos:
ill look into it quickly. may be easy

hYdos:
ok i just checked its probably going to be in my next pr when i add a fuel loader model and texture

PR #48, opened 2020-07-03
New manufactured item textures (battery -> oxygen concentrator)

ITotallyExist:
Replaced the textures for the basic and advanced wafer.


craftablescience:
All of these textures look great, but can you scale them down to 16x16 wherever possible? It goes better with the textures of the mod and vanilla Minecraft.

Issue #49, opened 2020-07-16
Shift clicking oxygen mask into slot hardlocks the game.


TheMCSpacePro:
Galacticraft Rewoven version: 0.2.5-alpha+1.16-pre5

Singleplayer, Multiplayer, or Singleplayer opened to LAN? Singleplayer

Please list the Galacticraft Rewoven addons you have installed
N/A
Please provide your LOG (preferably on [gist](https://gist.github.com) or [pastebin](https://pastebin.com)) and screenshots etc. if applicable
https://paste.ee/p/tm45W   (From what I can tell, the error is at line 1809, or https://paste.ee/p/tm45W#s=0&l=1809)

TheMCSpacePro:
Ok, accoring to the Mod Menu ingame, the version is v0.10-alpha+build.0. I am also on the 7/15/2020 build of the code.

craftablescience:
That's the version for every compiled version of the mod. Try redownloading the project and running it again.

marcus8448:
This bug occurs in (almost) all of our inventories. When slots are locked down to specific items, it causes an infinite loop in `ScreenHandler#method_30010`. I haven't had the time to look into it, so if someone wants to, please do.

craftablescience:
At least it doesn't seem to crash the game, and shift-clicking still works as intended.

PR #50, opened 2020-07-29
Fixed shift-clicking problem (fixes issue #49)

BeetMacol:
It is caused by that loop in `ScreenHandler#method_30010`:
`              for(itemStack7 = this.transferSlot(playerEntity, i); !itemStack7.isEmpty() && ItemStack.areItemsEqualIgnoreDamage(slot4.getStack(), itemStack7); itemStack7 = this.transferSlot(playerEntity, i)) {
                  itemStack = itemStack7.copy();
               }`

When i shift clicked iron boots it called `PlayerScreenHandler#transferSlot`, but when i did it with galacticraft tank, it called `ScreenHandler#transferSlot` because `this` (the ScreenHandler) is not an instance of `PlayerScreenHandler` but an instance of `PlayerInventoryGCScreenHandler`. I solved the problem by adding the transferSlot function to `PlayerInventoryGCScreenHandler`.

### DIFF
--- a/src/main/java/com/hrznstudio/galacticraft/screen/PlayerInventoryGCScreenHandler.java
+++ b/src/main/java/com/hrznstudio/galacticraft/screen/PlayerInventoryGCScreenHandler.java
@@ -177,9 +177,12 @@ public class PlayerInventoryGCScreenHandler extends ScreenHandler {
         if (slotFrom != null && slotFrom.hasStack()) {
             ItemStack stackFrom = slotFrom.getStack();
             stack = stackFrom.copy();
-            EquipmentSlot preferredEquipmentSlot = MobEntity.getPreferredEquipmentSlot(stack);
-            if ((index >= 0 && index < 8)) {
+            if (index >= 0 && index < 8) {
                 if (!this.insertItem(stackFrom, 9, 48, true)) {
+                    return ItemStack.EMPTY;
+                }
+            } else if (index >= 9 && index < 49) {
+                if (!this.insertItem(stackFrom, 0, 8, true)) {
                     return ItemStack.EMPTY;
                 }
             }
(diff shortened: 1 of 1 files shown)


src/main/java/com/hrznstudio/galacticraft/screen/PlayerInventoryGCScreenHandler.java
```diff
@@ -169,4 +170,41 @@ public int getMaxStackAmount() {
             return Pair.of(PlayerScreenHandler.BLOCK_ATLAS_TEXTURE, new Identifier(Constants.MOD_ID, Constants.SlotSprites.OXYGEN_TANK));
         }
     }
+
+    public ItemStack transferSlot(PlayerEntity player, int index) {
+        ItemStack stack = ItemStack.EMPTY;
+        Slot slotFrom = (Slot)this.slots.get(index);
```
marcus8448, line 176: I assume the slots list contains `Slot`s so the cast is unnecessary.
```suggestion
        Slot slotFrom = this.slots.get(index);
```

src/main/java/com/hrznstudio/galacticraft/screen/PlayerInventoryGCScreenHandler.java
```diff
@@ -169,4 +170,41 @@ public int getMaxStackAmount() {
...
+        if (slotFrom != null && slotFrom.hasStack()) {
+            ItemStack stackFrom = slotFrom.getStack();
+            stack = stackFrom.copy();
+            if (index >= 0 && index < 8) {
```
marcus8448, line 180: Index should always be greater than 0 as `slots.get(index)` would throw an exception if it wan't
```suggestion
            if (index < 8) {
```

src/main/java/com/hrznstudio/galacticraft/screen/PlayerInventoryGCScreenHandler.java
```diff
@@ -31,6 +31,7 @@
 import com.hrznstudio.galacticraft.screen.slot.ItemSpecificSlot;
 import com.mojang.datafixers.util.Pair;
 import net.minecraft.entity.EquipmentSlot;
+import net.minecraft.entity.mob.MobEntity;
```
marcus8448, line 34: Unused import?
```suggestion
```

PR #57, opened 2020-08-14
Update to 1.16.2

marcus8448:
A bunch of stuff.

- machines no longer use `BlockEntityRenderers` and instead use a runtime generated `FabricBakedModel`
- adds the bubble distributor
- adds structures to the moon
    - moon villages actually generate and have structures now
- moon villager model has been 'fixed'
- moon biomes have been reworked
- moon chunk gen has also been changed
- oxygen is no longer an 'energy type', and has become a fluid


craftablescience:
data/gcr/structures/moon_village/misc/sapings should be saplings

src/main/java/com/hrznstudio/galacticraft/api/wire/WireConnectionType.java
```diff
@@ -23,14 +23,17 @@
...
+ */
 public enum WireConnectionType {
     /**
      * The wire is not connected to anything.
      */
     NONE,
 
     /**
-     * The wire is connected to another wire.
+     * The wire is connected to another wiRE.
```
JoeZwet, line 36: why?

PR #59, opened 2020-08-26
Worldgen v3

marcus8448:
Structures are now registered using the new structures api


JoeZwet (approved):
this looks really good, you should show off some screenshots in fabricord

PR #64, opened 2020-09-14
Major machine refactors

marcus8448:
Side config data is now stored in the `BlockEntity` rather than `BlockState`. Saves us 3 seconds of loading time and *many, many* `BlockState` permutations.
I've redone wires yet again. The old implementation was incompatible with all other mods, and i used an inefficient data structure for the task.
  * wires now transfer *all* energy to the *nearest* machine, rather than splitting it up between all machines, no matter the distance. (Should I revert this change?)
Added block faces for item I/O (hoppers can properly interact with machines now)
Made `MachineHandledScreen` less of a cursed magic number mess


JoeZwet (approved):
looks good to me. not sure about the nearest machine thing for the wires but I guess it makes sense.

PR #69, opened 2020-10-14
change: minify JSON on build

marcus8448:


JoeZwet (approved):
very nice didn't even know you could do this.

PR #73, opened 2021-01-03
change: more machine refactors

marcus8448:
Updates minecraft to 1.16.4
Large changes, mainly in networking and GUIs.


JoeZwet (approved):
looks good.

not sure why we need to section out maven repo things in the build.gradle but ill approve anyway

marcus8448:
I configured the repos like that for speed - it makes it so that we only query the one repo for every dependency, rather than all of them (worst case).

PR #74, opened 2021-01-07
feat: lunar structures

marcus8448:
Dungeons are still NYI and will come in a future PR
pillager base exterior
pillager base inside example
lunar village
lunar village greenhouse
lunar ruins
not structure, just worldgen


JoeZwet (approved):
cheese.nbt

PR #75, opened 2021-01-24
cleanup: machine logic

marcus8448:
I've rewritten *all* machine logic to be consistent and have less duplicated code.


JoeZwet (approved):
my guy, you're insane.

PR #76, opened 2021-01-28
FIX: The /dimensiontp command now works properly

SebaSphere:
~~I fixed the implementation of you being teleported beneath bedrock when running the command and teleporting to the moon.~~

~~However, teleporting on the X and Z is broken.~~

This is now working without any problems.

### DIFF
--- a/src/main/java/com/hrznstudio/galacticraft/server/command/GalacticraftCommands.java
+++ b/src/main/java/com/hrznstudio/galacticraft/server/command/GalacticraftCommands.java
@@ -75,19 +75,18 @@ public class GalacticraftCommands {
         context.getSource().getMinecraftServer().execute(() -> {
             try {
                 ServerPlayerEntity player = context.getSource().getPlayer();
-                ServerWorld world = context.getSource().getWorld();
-
                 ServerWorld serverWorld = DimensionArgumentType.getDimensionArgument(context, "dimension");
                 if (serverWorld == null) {
                     context.getSource().sendError(new TranslatableText("commands.galacticraft-rewoven.dimensiontp.failure.dimension").setStyle(Style.EMPTY.withColor(Formatting.RED)));
                     return;
                 }
+                player.teleport(serverWorld,
+                        player.getX(),
+                        getTopBlockY(serverWorld, player),
+                        player.getZ(),
+                        player.yaw,
+                        player.pitch);
 
-                player.moveToWorld(serverWorld);
-                player.teleport(player.getX(),
-                        getTopBlockY(world, player),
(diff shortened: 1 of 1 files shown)


src/main/java/com/hrznstudio/galacticraft/server/command/GalacticraftCommands.java
```diff
@@ -104,4 +117,29 @@ private static int teleportMultiple(CommandContext<ServerCommandSource> context)
...
+     */
+    private static double getTopBlockY(ServerWorld world, ServerPlayerEntity player) {
+        int playerX = (int) player.getX();
+        int playerZ = (int) player.getZ();
+
+        for (int i = world.getHeight(); i > 0; i-- ) {
+            BlockPos pos = new BlockPos(new Vec3d(playerX, i, playerZ));
+            Block currentBlock = world.getBlockState(pos).getBlock();
+            if (currentBlock != Blocks.VOID_AIR) {
+                if (currentBlock != Blocks.AIR) {
```
marcus8448, line 136: Maybe use 'Block#isAir' (or is it a block state method?)

SebaSphere: Oh whoops, meant to keep that on the same line with a || operator. 
VOID_AIR would be any space that's not buildable, at least what I understood.

It works tho, however keeping it on one line could be useful to cleanup.

craftablescience: If you're going to still use that, night as well add cave air to the check just in case

src/main/java/com/hrznstudio/galacticraft/server/command/GalacticraftCommands.java
```diff
@@ -67,13 +75,19 @@ public static void register() {
...
                 }
-
-                context.getSource().getPlayer().moveToWorld(serverWorld);
+                player.teleport(serverWorld,
+                        player.getX(),
+                        getTopBlockY(serverWorld, player),
+                        player.getZ(),
+                        player.yaw,
+                        player.pitch);
+                System.out.println("HEY");
```
marcus8448, line 90: Remove debug log.

marcus8448 (changes requested):

marcus8448 (approved):

src/main/java/com/hrznstudio/galacticraft/server/command/GalacticraftCommands.java
SebaSphere: > If you're going to still use that, might as well add cave air to the check just in case

Marcus suggested using BlockState::isAir, so switched to that.

PR #78, opened 2021-01-29
fix: don't load client-only classes on the server

marcus8448:


craftablescience:
The problem with the particles is I was stupid enough to make an empty `register()` method in `GalacticraftParticles` and call it in the `Galacticraft` class. If you just remove line 100 in `Galacticraft` class, and then delete the `register()` method in `GalacticraftParticles`, the particles should be fixed, and it would keep with the pattern of registering stuff within helper methods. You could also move the `particle` package to the `client` package if you wanted, since it's all client side.

craftablescience:
The reason I made a helper class is because I thought we would need to make more particles in future -- GC has its fair share of them

craftablescience (approved):
👍

PR #79, opened 2021-01-31
Add Item Gravity

SteveKunG:
Just got some spare free time to help @craftablescience.


craftablescience:
I think this should be in the Addon-API repo, but this is genius
Thanks for saving me

SteveKunG:
Oh okay, will do!

PR #95, opened 2021-03-20
fix: Make it compile on certain IDEs again

Geolykt:
For some reason sometimes plain `File("src/main/resources/galacticraft_rewoven.accesswidener")` will result in the file being searched in a random directory (I think it was `home/Geolykt/.gradle/deamon/6.8.3/src/main/resources/galacticraft_rewoven.accesswidener` for me)

(please ignore that I accidentally deleted the branch)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -66,7 +66,7 @@ base {
 
 loom {
     refmapName = "galacticraft-rewoven.refmap.json"
-    accessWidener = File("src/main/resources/galacticraft_rewoven.accesswidener")
+    accessWidener = rootProject.file("src/main/resources/galacticraft_rewoven.accesswidener")
 }
 
 repositories {


build.gradle.kts
```diff
@@ -66,7 +66,7 @@ base {
 
 loom {
     refmapName = "galacticraft-rewoven.refmap.json"
-    accessWidener = File("src/main/resources/galacticraft_rewoven.accesswidener")
+    accessWidener = File(project.buildDir.toString() + "/../src/main/resources/galacticraft_rewoven.accesswidener")
```
marcus8448, line 69: Can you test if this works for you? (I'd like to avoid the string concatenation)
```suggestion
    accessWidener = rootProject.file("src/main/resources/galacticraft_rewoven.accesswidener")
```

Geolykt: It should and it does

JoeZwet (approved):
good catch

PR #97, opened 2021-04-03
fix: mixin remapping and targetting

marcus8448:
Also adds a mixin plugin so that we don't apply these debug mixins outside of the development environment
Removes an obsolete accessor (the constructor for `BlockStateProviderType` is `public` now)


JoeZwet (approved):
:+1:

PR #99, opened 2021-04-04
feat: add version decoration

JoeZwet:
image
adds `+build.<branch>.<commit>` to the version.
The branch is not included if it is `main`
This allows for better debugging when players come in with their self-compiled versions asking why something isn't working.\
Set the `RELEASE` environment var to build without the decoration.


build.gradle.kts
```diff
@@ -250,3 +250,32 @@ license {
...
     }
 }
+
+// inspired by https://github.com/TerraformersMC/GradleScripts/blob/2.0/ferry.gradle
+fun getVersionDecoration(): String {
+    if(project.hasProperty("release")) return ""
+
+    var version = "+build"
+    val branch = "git branch --show-current".execute()
+    if(branch.isNotEmpty() && branch != "main") {
```
marcus8448, line 260:
```suggestion
    if(branch.isNotEmpty() && branch.equals("main")) {
```

marcus8448 (changes requested):
I'm assuming that == and #equals works the same as java.

build.gradle.kts
JoeZwet: wouldnt it be `!branch.equals("main")`?

JoeZwet:
current build error:

```
FAILURE: Build failed with an exception.

* Where:
Build file '/workspace/build.gradle.kts' line: 277

* What went wrong:
Cannot run program "git" (in directory "/workspace"): error=2, No such file or directory
> error=2, No such file or directory
```
trying to find a fix

build.gradle.kts
JoeZwet: tho intellij seems to think that `!=` is better 
image

marcus8448: Ok, I assume kotlins equals handling is different than java then.

marcus8448 (commented):
Should there be an override for whatever build system we'll be using? If we don't have a build number before the `+` no version parser will be able to tell if `d72dc71` is newer than `f4eb290`.

build.gradle.kts
```diff
@@ -250,3 +250,33 @@ license {
...
+// from https://discuss.gradle.org/t/how-to-run-execute-string-as-a-shell-command-in-kotlin-dsl/32235/5
+fun String.execute(workingDir: File = project.file("./")): String {
+    val parts = this.split("\\s".toRegex())
+    val proc = ProcessBuilder(*parts.toTypedArray())
+        .directory(workingDir)
+        .redirectOutput(ProcessBuilder.Redirect.PIPE)
+        .redirectError(ProcessBuilder.Redirect.PIPE)
+        .start()
+
+    proc.waitFor(1, TimeUnit.MINUTES)
```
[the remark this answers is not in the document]
JoeZwet: iirc just outputs as is

marcus8448 (approved):

PR #102, opened 2021-04-12
Feat: Pipe Walkway, Optimization

craftablescience:
image

- Drops the walkway attachment outlines, they were clogging up memory and didn't need to be a part of the outline regardless. Now uses a 2^6 (64) large cache instead of 2^12 (4096) for each walkway.
- Pipe Walkway added, has the same features as the Wire Walkway but it's a pipe. It will need to be updated along with the fluid pipes, but inheritance should take care of most of the changes made to pipes.
- Removed the old Python program I made to speed up BlockBench entity model creation, now that BlockBench can export to Fabric it was redundant.
- Refactored the glass pipe blockstate (~4,000 -> ~200 lines).
- Glass pipes have nicer looking colors.
- Reduced complexity of general walkway code.
- Added sealable aluminum wire texture to remove error on load.
- Removed deleted texture in refinery JSON blockmodel to remove error on load.
- Wire and Pipe Walkways connect on their top face to fix networks connecting on invisible connections.

**Known bug:** placing a block that extends FluidLoggableBlock will crash the game. This is also present on main. I removed the fluid loggable capability from the Pipe Walkway because it might create weird issues depending on how the pipes will eventually render, but the other walkways are still loggable.


src/main/java/com/hrznstudio/galacticraft/block/special/walkway/PipeWalkway.java
```diff
@@ -0,0 +1,276 @@
...
+                i |= getFacingMask(Direction.SOUTH);
+            if (blockState.get(FACING).equals(Direction.EAST))
+                i |= getFacingMask(Direction.EAST);
+            if (blockState.get(FACING).equals(Direction.WEST))
+                i |= getFacingMask(Direction.WEST);
+            if (blockState.get(FACING).equals(Direction.UP))
+                i |= getFacingMask(Direction.UP);
+            if (blockState.get(FACING).equals(Direction.DOWN))
+                i |= getFacingMask(Direction.DOWN);
+            return i;
```
marcus8448, line 162:
```suggestion
            return getFacingMask(blockState.get(FACING));
```
I don't totally understand what's happening here, but I think the cache can just be Direction -> int, since that's the only thing where ted about the state.

Probably the same thing with the other walkway classes.

src/main/java/com/hrznstudio/galacticraft/block/special/walkway/WireWalkway.java
```diff
@@ -251,7 +215,7 @@ public boolean canConnect(BlockState state, BlockState neighborState, BlockPos p
                 return false;
         } catch (IllegalArgumentException ignored) {}
         // TODO: The WireBlockEntity will still connect on the top face of this block (there's no wire there)
```
marcus8448, line 253: Has this been fixed?

marcus8448 (commented):
I'll take a closer look later today

src/main/java/com/hrznstudio/galacticraft/block/special/walkway/PipeWalkway.java
craftablescience: You're probably right, I was just simplifying the old code where it also used the connections to the block.

src/main/java/com/hrznstudio/galacticraft/block/special/walkway/PipeWalkway.java
```diff
@@ -0,0 +1,262 @@
...
+public class PipeWalkway extends FluidPipe {
+    public static final BooleanProperty NORTH = Properties.NORTH;
+    public static final BooleanProperty EAST = Properties.EAST;
+    public static final BooleanProperty SOUTH = Properties.SOUTH;
+    public static final BooleanProperty WEST = Properties.WEST;
+    public static final BooleanProperty UP = Properties.UP;
+    public static final BooleanProperty DOWN = Properties.DOWN;
+    public static final DirectionProperty FACING = Properties.FACING;
+    private static final VoxelShape[] shape = new VoxelShape[64];
+    private final Object2IntMap<BlockState> SHAPE_INDEX_CACHE = new Object2IntOpenHashMap<>();
```
marcus8448, line 65: This should be static.
Also, since it only checks the direction, make it an `Object2IntMap<Direcition>` so that there are less things to cache? (Do we even need to cache it? It only calculates `1 << direction.getId()`)

craftablescience: To be entirely honest I copied this from the aluminum wire class half a year ago, I've forgotten its purpose by now. I'll see if I can get rid of it

marcus8448 (approved):
LGTM

PR #116, opened 2021-04-26
feat: additional textures

Pomi108:


craftablescience (changes requested):
I think most minecraft buckets have drops going down the sides, but otherwise I like all the textures here :)

Pomik108:
actually it's only the lava bucket and the fish buckets. The general rule seems to be that if you dont want to spill it, it doesnt have drops – and you dont want to spill oil nor fuel, so i went with no drops

craftablescience:
The powder buckets have it as well, I think it's a bit of an oversight they didn't add it to the water bucket personally

Pomik108:
it might be the case.

JoeZwet:
Please be sure to read our [contributing guidelines](https://github.com/TeamGalacticraft/Galacticraft/blob/main/.github/CONTRIBUTING.md), specifically the section on commit names.

Pomik108:
Whoops my bad, didn't know these were a thing. I'll make sure all my future commits follow these conventions

vos6434:
Our artists will remaster the textures.

PR #117, opened 2021-04-27
change: moon world generation

marcus8448:
* change: make mare biomes rarer
* change: make biomes larger
* change: make edge biomes generate properly

Cleans up moon worldgen so that if someone wants to implement another planet, its easier to copy


craftablescience (approved):
That json though
```
[

]
```
Right in the middle of all the closed brackets `[]`
🤣

PR #122, opened 2021-05-05
feat: add pr spec to contributing guidelines

craftablescience:
Yes, I am aware of the irony of making a PR to add 2 lines to a text file after just saying that isn't a good idea in the changes to the text file.

Any thoughts on this?


JoeZwet:
Maybe add that pr titles should follow the commit spec. I don't think it's something we ever discussed but we all just started doing so it might be worth adding.

craftablescience:
Yeah, that's a good idea.

JoeZwet (approved):

PR #125, opened 2021-05-14
feat: improve Github actions

JoeZwet:


.github/workflows/build.yml
```diff
@@ -1,24 +1,43 @@
...
   build:
-    runs-on: ubuntu-20.04
-    container:
-      image: openjdk:8-jdk
+    strategy:
+      matrix:
+        java: [
+          1.8, # Minimum
+          11,  # LTS
+          16,  # Latest
```
marcus8448, line 14: Do we really need to test on all these versions? Minimum should be good enough, or even minimum and latest, if there's no forward compatibility for some reason.

.github/workflows/build.yml
```diff
@@ -1,24 +1,43 @@
...
+        run: chmod +x ./gradlew
+      - name: Build
         uses: eskatos/gradle-command-action@v1
         with:
           arguments: checkLicenses build
           wrapper-cache-enabled: true
           dependencies-cache-enabled: true
-      - name: capture build artifacts
+      - name: Capture build artifacts
+        if: ${{ runner.os == 'Linux' && matrix.java == '11' }}
```
marcus8448, line 39: Maybe capture artifacts for 8 - there *technically* is a backwards compatibility issue with `FloatBuffer#flip` (signature changed [but also somehow forwards compatible]), not that we're using that yet.

.github/workflows/pr_check.yml
```diff
@@ -1,19 +1,34 @@
...
-      - name: validate gradle wrapper
+        with:
+          fetch-depth: 0
+      - name: Validate Gradle wrapper
         uses: gradle/wrapper-validation-action@v1
-      - name: build
+      - name: Setup JDK ${{ matrix.java }}
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java }}
```
marcus8448, line 27:
```suggestion
        uses: actions/setup-java@v2
        with:
          java-version: ${{ matrix.java }}
          distribution: 'adopt'
```

.github/workflows/build.yml
```diff
@@ -1,24 +1,43 @@
...
-      - name: validate gradle wrapper
+        with:
+          fetch-depth: 0
+      - name: Validate Gradle wrapper
         uses: gradle/wrapper-validation-action@v1
-      - name: build
+      - name: Setup JDK ${{ matrix.java }}
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java }}
```
marcus8448, line 28:
```suggestion
        uses: actions/setup-java@v2
        with:
          java-version: ${{ matrix.java }}
          distribution: 'adopt'
```

.github/workflows/pr_check.yml
```diff
@@ -1,19 +1,34 @@
...
 
 jobs:
   build:
-    runs-on: ubuntu-20.04
-    container:
-      image: openjdk:8-jdk
+    strategy:
+      matrix:
+        java: [
+            1.8, # Minimum
```
marcus8448, line 11:
```suggestion
            8, # Minimum
```
Required for `actions/setup-java@v2`

.github/workflows/build.yml
```diff
@@ -1,24 +1,43 @@
...
 
 jobs:
   build:
-    runs-on: ubuntu-20.04
-    container:
-      image: openjdk:8-jdk
+    strategy:
+      matrix:
+        java: [
+          1.8, # Minimum
```
marcus8448, line 12:
```suggestion
          8, # Minimum
```
Required for `actions/setup-java@v2`

.github/workflows/publish.yml
```diff
@@ -2,20 +2,28 @@ name: Maven Publish
...
+      - name: Checkout repository
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Validate Gradle wrapper
+        uses: gradle/wrapper-validation-action@v1
+      - name: Setup JDK 1.8
+        uses: actions/setup-java@v1
         with:
           java-version: 1.8
```
marcus8448, line 20:
```suggestion
        uses: actions/setup-java@v2
        with:
          java-version: 8
          distribution: 'adopt'
```

marcus8448 (changes requested):

.github/workflows/build.yml
JoeZwet: iirc most people will use lts for custom installs ie multimc

.github/workflows/build.yml
```diff
@@ -1,24 +1,44 @@
...
-    container:
-      image: openjdk:8-jdk
+    strategy:
+      matrix:
+        java: [
+          8, # Minimum
+          11,  # LTS
+          16,  # Latest
+        ]
+        os: [ubuntu-20.04]
```
marcus8448, line 16: There's no reason to have this if the array only contains one entry. (Isn't java meant to be 'write once, run anywhere' anyways?)

.github/workflows/build.yml
```diff
@@ -1,24 +1,44 @@
...
+      - name: Validate Gradle wrapper
         uses: gradle/wrapper-validation-action@v1
-      - name: build
+      - name: Setup JDK ${{ matrix.java }}
+        uses: actions/setup-java@v2
+        with:
+          java-version: ${{ matrix.java }}
+          distribution: 'adopt'
+      - name: Make Gradle wrapper executable
+        if: ${{ runner.os != 'Windows' }}
```
marcus8448, line 31: We never actually run on windows

.github/workflows/build.yml
JoeZwet: resolved in c02a824

.github/workflows/build.yml
```diff
@@ -1,24 +1,44 @@
...
+        run: chmod +x ./gradlew
+      - name: Build
         uses: eskatos/gradle-command-action@v1
         with:
           arguments: checkLicenses build
           wrapper-cache-enabled: true
           dependencies-cache-enabled: true
-      - name: capture build artifacts
+      - name: Capture build artifacts
+        if: ${{ runner.os == 'Linux' && matrix.java == '11' }}
```
marcus8448: We always run on linux

.github/workflows/pr_check.yml
```diff
@@ -1,21 +1,37 @@
...
-    container:
-      image: openjdk:8-jdk
+    strategy:
+      matrix:
+        java: [
+            8, # Minimum
+            11,  # LTS
+            16,  # Latest
+        ]
+        os: [ubuntu-20.04]
```
marcus8448, line 15: Same thing - There's no reason to have this if we're only using ubuntu.

.github/workflows/pr_check.yml
```diff
@@ -1,21 +1,37 @@
...
+      - name: Validate Gradle wrapper
         uses: gradle/wrapper-validation-action@v1
-      - name: build
+      - name: Setup JDK ${{ matrix.java }}
+        uses: actions/setup-java@v2
+        with:
+          java-version: ${{ matrix.java }}
+          distribution: 'adopt'
+      - name: Make Gradle wrapper executable
+        if: ${{ runner.os != 'Windows' }}
```
marcus8448, line 30: We don't run actions on Windows.

.github/workflows/publish.yml
```diff
@@ -2,20 +2,29 @@ name: Maven Publish
...
+          fetch-depth: 0
+      - name: Validate Gradle wrapper
+        uses: gradle/wrapper-validation-action@v1
+      - name: Setup JDK 1.8
+        uses: actions/setup-java@v2
+        with:
+          java-version: 8
+          distribution: 'adopt'
+      - name: Make Gradle wrapper executable
+        if: ${{ runner.os != 'Windows' }}
```
marcus8448, line 23: We don't run this on Windows (this action doesn't have a os matrix either)

marcus8448 (changes requested):

.github/workflows/publish.yml
JoeZwet: its more of a sanity check for if we ever do

.github/workflows/build.yml
marcus8448: What I meant is: Is it necessary to test compilation on all three versions? Java *shouldn't* be breaking old things with new releases (Unless we're using unsupported APIs like `Unsafe`).

.github/workflows/build.yml
JoeZwet: yeah, but this is just future proofing it, in case we ever do want to build on windows its already set up to do so.

.github/workflows/build.yml
JoeZwet: sanity check for if we ever do

.github/workflows/build.yml
```diff
@@ -1,24 +1,44 @@
...
         with:
           arguments: checkLicenses build
           wrapper-cache-enabled: true
           dependencies-cache-enabled: true
-      - name: capture build artifacts
+      - name: Capture build artifacts
+        if: ${{ runner.os == 'Linux' && matrix.java == '8' }}
         uses: actions/upload-artifact@v2
         with:
           name: Artifacts
```
marcus8448, line 43:
```suggestion
          name: Galacticraft
```

.github/workflows/build.yml
```diff
@@ -1,24 +1,44 @@
...
-          path: build/libs/
+          path: build/libs
```
marcus8448, line 44:
```suggestion
          path: |
            build/libs/
            !build/libs/*-sources.jar
            !build/libs/*-dev.jar
```

marcus8448 (changes requested):
Might as well also toss this in

.github/workflows/build.yml
JoeZwet: it's probably not needed but it's a good sanity check so we know it works on multiple versions.

marcus8448 (approved):

PR #156, opened 2021-09-24
chore: defer registration of objects to mod init

marcus8448:
Prevents some annoying classloading order bug(s) from occurring again.


grantlmul:
looks nicer too imo

PyrrhaDevs (approved):
👍

PR #166, opened 2021-10-01
fix: thermal leggings pattern + recipes

mMONTAGEe:
Added recipes for Nuggets ,Advanced Solar Panel  and Dark Decoration (uses desh)

PR #167, opened 2021-10-01
fix/chore: fix launch pad and merge changes from main

grantlmul:
I've merged main to feature/rockets, making (hopefully the right) edits to have things go over well and follow the same style.
I also have fixed the launch pad model and renamed `launch_pad` to `rocket_launch_pad`. If this needs to go the other way around instead, I'll change that.

In `screen.GalacticraftScreenHandlerType`, I've noticed that things for rocket related stuff gets registered a bit differently, and I'm not sure if/what I need to change to have it conform with everything else registered in the style of changes made in 755fdfc.


grantlmul:
oh hello #166 :))))

grantlmul:
gg ez

grantlmul:
i think i misclicked something

Issue #161, opened 2021-09-26
Attention Required


mMONTAGEe:
Here is my list. !Do not close it as i will slowly fill it!
================================
**BLOCKS N GENERAL** (needs verifications)

-Lunar Cartographics Table GUI won't show up.

-Electric Compressor side configuration won't work

-Advanced Solar Panel faces wrong/ does not face the star at all.

-frequency module sometimes can randomly crash your game. or when full oxygen complect is on you

"-Refinery does not refine anything" -> About to be fixed by marcus

-Most machines does nothing

-Space helmet wiggles like on drugs

-Little zombie have wrong mask position

=========================
**TEXTURE ISSUES**

-Spider have broken textures

-Creeper  have broken textures

=============================
**MOON**

-Moon sunset have atmospheric yellow colors

==============================
**MOON VILLAGE**

-Village generations should be corrected.

-Moon Villagers immediatly die in space. that means they need to replaced with Alien guys.

==============================
**OTHER**

-Single Solar Module and Full Solar Panel recipes doesn't work and won't show up in JEI

**Those are common things that i found. Further analysis will reveal more**

PyrrhaDevs:
Hey just for future reference: please use one issue per bug. It makes it a lot easier for us to track what's fixed and what's not

mMONTAGEe:
issues with blocks need to reverify with latest changes and update this sheet

PR #179, opened 2021-10-08
fix: compressor and fabricator output

ericb147:
compressor and fabricator now output correctly. 

this addresses some issues from #161


ericb147 (commented):
Changes from Marcus

PR #182, opened 2021-10-14
feat: discord notify and test setup

JustPyrrha:
This fixes the current failing build on `main` due to the old `TeamGalacticraft/discord-build-notify` was deleted without being removed from the `build.yml` workflow.

It also sets up proper test reporting from GameTest and adds the basics for JUnit5 testing.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,7 @@ jobs:
         uses: actions/setup-java@v2
         with:
           java-version: 16
-          distribution: temurin 
+          distribution: temurin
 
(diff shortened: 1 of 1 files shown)


marcus8448:
Could you also copy over the relevant changes (`gradle/gradle-build-action`, jdk distribution) to the `codeql` and `publish` workflows?

PyrrhaDevs:
> Could you also copy over the relevant changes (`gradle/gradle-build-action`, jdk distribution) to the `codeql` and `publish` workflows?

done in `c63bd14`

PR #184, opened 2021-10-15
fix: part of pipe_walkway texture not working

Domza64:
Both pipe and walkway textures now work for pipe_walkway item.
Before pipe texture in pipe_walkway was pink/black thing.


Domza64:
I just fixed texture UVs, it now looks how it is supposed to look.

PR #187, opened 2021-10-19
chore: machine tests

marcus8448:
does not include solar panels, oxygen-based machines
also fixes gametest source set creation


PyrrhaDevs:
so on that, you'd have a test file for just say Wire tests and another just for pipe tests and these files would contain all the separate test for that one block/structure/etc

PR #192, opened 2021-10-27
Fix: cape json loading

### LINKED ISSUE #191: Capes Issue
### Mod Loader

Fabric

### Version Information

none, just looking

### Log or Crash Report

none

### Reproduction steps

I've noticed that the capes only count as only one object, Loaded capes for 1 players. (Took 373ms)

the reason for this is that it is not looking into the "players" object

it should be an easy fix

craftablescience:
Fixes #191.

If you would like to test it, replace the URL in `CapesLoader.java` with `https://raw.githubusercontent.com/TeamGalacticraft/Galacticraft/fix/cape-json-loading/capes.json`.


PyrrhaDevs (approved):
lang changes are out of scope but its fine

craftablescience:
> 
> 
> lang changes are out of scope but its fine

sorry, it was really bugging me 😅

PR #124, opened 2021-05-14
Feature: Added some blocks/items and bug fixes

SteveKunG:
### Blocks :
* Moon Basalt Copper Ore
* Bacterial Sludge
* Fallen Meteor

### Items :
* Atomic Battery
* Ambient Thermal Controller

### Bug Fixes :
* Fixed class cast exception when placing wire near machines
* Fixed Wire and Glass Pipe slow connected visual update
* Fixed using dye on Glass Pipe consumed in Creative Mode
* Fixed Cavernous Vines block rendering
* Fixed Fabrication recipes not working
* Fixed wrong registry name in most recipes
* Fixed Tin Ladder item consumed if it can't place
* Fixed Tin Ladder placed sound
* Fixed SoundEvent crashed on server side

### Asset Fixes :
* Added missing item model for Sealable Aluminum Wire
* Added missing translation key for Moon Surface Rock
* Fixed Desh Leggings using Desh Boots texture
* Move out Galacticraft Armor textures from 'minecraft' folder

### Misc :
* Ignore Eclipse files
* Minor changes to Solar Part block shape
* Clean up connected block states name
* Fix inconsistence name for some blocks/items
* Clean up FluidBlock class
* Made all blocks require specific tool to mine

PR #206, opened 2021-11-22
Feat: Added Fallen Meteor

SteveKunG:
Seperate from #124 

~My current fix concept is made `HeatLevel` into 0-4 `heat` block state. This should be better than updating block color every tick.~

~Still has bad performance when multiple fallen meteors are in one chunk.~ Fixed!


DarianLStephens:
Are they still block entities?
I feel like it would be better if they were regular blocks with differently-colored models for the states that just updated heat level on weighted random tick, since it barely changes anything.
FPS will probably still struggle with so many hot meteors in a chunk because it's updating all that complex geometry so often, but once they cool down, there should be next to no additional performance impact.

SteveKunG:
@DarianLStephens Yeah, I'm going to phase it out.

This is actually good so far.

SteveKunG:
Testing with both FPS and TPS are fine.

AlphaMode (approved):

DarianLStephens:
Eyy, it finally got merged!

PR #241, opened 2022-10-24
feat: update to minecraft 1.19.2

marcus8448:
- Drops the use of LBA and replaces it with [MachineLib](https://github.com/TeamGalacticraft/MachineLib).
  - A bunch of machine apis have moved to that project
- Replaces yarn with Mojang/official mappings
- Initial data generation setup
- Changes to moon worldgen

What needs to be done:
- [x] Stabilize GCAPI 0.4
- [x] Rename Galacticraft classes/methods/etc to match the official mappings


JustPyrrha:
the switch to official mappings will be huge for when we start supporting more loaders, good shit

AlphaMode (approved):

PR #244, opened 2022-11-25
update models of archgrey, grey, commetcube for animation

Nimu93:


JustPyrrha (changes requested):
please make sure you've read our contributing guidelines as it outlines things such as our commit message standard

PR #250, opened 2022-12-20
Merge Galacticraft API into galacticraft as a gradle sub project.

AlphaMode:
This PR merges [GalacticraftAPI](https://github.com/TeamGalacticraft/GalacticraftAPI) into Galacticraft as a gradle sub project.
This makes compiling the mod easier, and helps with updating and maintaining the api.


AlphaMode:
Scraped for now. Although maybe possibly look into this in the future...

PR #251, opened 2023-01-14
Enhances Newcomer Experience

T-A-B:
This took a lot of fiddling around to figure out... so it may help others in the future by including something like this in the readme.  The location that I chose is arbitrary... but something like this would surely be useful.


supersimple33 (approved):

AlphaMode:
This is outdated now

Issue #255, opened 2023-01-25
Game crash on /give player rocket


supersimple33:
### Mod Loader

Fabric

### Version Information

7d5137c

Reconfirmed in 13285bb

### Log or Crash Report

https://gist.github.com/supersimple33/f5ec952b49319ca606d7b48ef44b311a

### Reproduction steps

When running `/give @p galacticraft:rocket` the game crashes. I believe it could have something to do with the rotating texture of the item.

supersimple33:
Appears as though this is expected for now as the problem is from a nbt tag which for the rocket which is null since they have not been implemented in the API. As a workaround a null check could be added in `RocketDataImpl` but I think such a temporary work around is unnecessary.

I'll leave the issue open since it is technically still a problem.

supersimple33:
Fixes confirmed

PR #256, opened 2023-01-25
feat: Adding the nasa workbench

supersimple33:
I have added the nasa workbench. To do this I have done the following:

- Added and registered the NASA workbench block and gave it a crafting recipe.
- Built on the `NasaWorkbenchBlock` to handle being opened and sending the user to the correct menu screens
- Create a class structuring for the Menus of the device
  - The `AbstractNasaWorkbenchMenu` class is to be extended by rocket schematic classes so that they only need to register where their crafting slots are for the different schematics
  - Since it is not a crafting menu the `NasaWorkbenchSchematicMenu` gets its own separate class which handles holding onto the schematic the user gives it
- Created class hierarchy for the Screens of the device
  - The abstract class `AbstractNasaWorkbenchScreen` handles the shared logic of all the menu screens which is mostly the rendering and adding the nav buttons on the left
  - The `NasaWorkbenchSchematicScreen` also has the unlock button
- To communicate what recipes the user knows I decided to use advancements since they are held server side but will allow the user to look at what recipes they know and maybe in the future we can add some that are not strictly for making rockets but ones that will help guide the player
- Finally I added a new Recipe the `RocketeeringRecipe` which allows recipes to be encoded for each rocket much the same way that all other recipes are made
  - Because of #255 I can't set the recipe to output `galacticraft:rocket` or else the game crashes so I have left it as sand for now but it is a simple one line change once fixed
- I also added a few other things to make the bench work such as the rocket parts items


supersimple33:
Ok seems this may need some changes didn't know of the modular rockets plan.

AlphaMode:
Creating rockets are still being planned out. We don't really plan on bringing back the nasa workbench, but this could work as a temporary solution until then...

src/main/java/dev/galacticraft/mod/Constant.java
```diff
@@ -712,6 +712,7 @@ interface ScreenHandler {
         String OXYGEN_SEALER_SCREEN_HANDLER = "oxygen_sealer_screen_handler";
         String FUEL_LOADER_SCREEN_HANDLER = "fuel_loader_screen_handler";
         String AIR_LOCK_CONTROLLER_MENU = "air_lock_menu";
+        String NASA_WORKBENCH_SCREEN_HANDLER = "nasa_workbench_screen_handler";
```
AlphaMode: The name should end with menu since we are following mojmap standards and not yarn standards anymore

AlphaMode (changes requested):

PR #258, opened 2023-01-29
fix: Added directional block state to fix wire textures

### LINKED ISSUE #253: Wire textures not updating properly
### Mod Loader

Fabric

### Version Information

7d5137c

### Log or Crash Report

No crash

### Reproduction steps

When placing wires and fluid pipes the textures will not update properly to match the connections between them, however the hit boxes are correct.

supersimple33:
I added in directional blockstates similar to how things look in the HeavyAluminumWire. This fixes #253 so that when you place wires down they look how they are supposed to.


supersimple33:
This has been resolved in the latest changes and pr is no longer needed.

PR #259, opened 2023-01-31
feat: Adding some recipes

supersimple33:
- Added a recipe for cryo bed following the legacy GC recipe
- Added recipes to go from nuggets to ingots for GC raw materials
- Moved duplicate recipes into single files to reduce space and keep code in the same place
- Added a recipe for silicon ore smelting for those with a silk touch pickaxe
- Removed recipes for basalt based ore since those ore do not exist

(Commits went burr in the last PR so I remade it)


AlphaMode (changes requested):
Recipes should be data generated and added via https://github.com/TeamGalacticraft/Galacticraft/blob/main/src/main/java/dev/galacticraft/mod/data/GCRecipeProvider.java

supersimple33:
@AlphaMode Addressed requested changes putting the new recipes into the `GCRecipeProvider.java`.
Also I added tags for tracking items with the same smelting output, and also added compatibility for blasting.

supersimple33:
Also, do we want to track src/main/generated/.cache/ or should I put it in the .gitignore

supersimple33:
I'm just going to remake this PR

Issue #252, opened 2023-01-18
Solar panels have wrong texture in default state


supersimple33:
### Mod Loader

Fabric

### Version Information

7d5137c1

Reconfirmed in 13285bb

### Log or Crash Report

No crash

### Reproduction steps

If you place either of the two solar panels one will notice that the texture for the back in the default state conveys energy out, however I believe it should be the hazard tape and not energy out texture in the default state since the default IO config will not output energy.

supersimple33:
Also seeing this on the coal generator and circuit fabricator

supersimple33:
This is fixed for solar panels but I do think there may be other machines that need to have their texture re-shuffled

PR #261, opened 2023-02-23
fix: default solar panel textures (#252)

Christoffyw:


supersimple33 (approved):
Looks good, confirmed this fixes #252.

PR #265, opened 2023-06-19
chore: Small cleanup changes

supersimple33:
If you clone the repo right now ./gradlew has lost executable permissions so I fixed that. And also gradlew will complain that a few files are missing licenses at the top so I fixed that too.


AlphaMode (approved):
LGTM

PR #273, opened 2023-07-12
hotfix: Registering fluid variants attributes after blocks

supersimple33:
I split apart the stages of registering in GCFluids so that fluidVariantAttributes can inherit the names when calling `GCBlocks.<block>.getDescriptionId()` while not requiring the GCFluids to be registered after GCBlocks.


supersimple33:
@SteveKunG this will not interfere with your solution correct?

SteveKunG:
Yeah, this looks good!

SteveKunG (approved):

PR #274, opened 2023-07-13
feat: Adding resource compression recipes

supersimple33:
I have added a few helper methods in the `GCRecipeProvider` class which assist with making 9x9 compression and decompression recipes. I have used these methods to transfer the transfer the recipes for metallic blocks, ingots, and items to the datagen. Also there used to not nugget -> ingot recipes even before they were moved to disabled which is why I originally made this feature branch.


supersimple33:
Build is fine locally its failing here since jfrog disabled free tier and thus build can't find [badpackets](https://landing.jfrog.com/reactivate-server/bai).

AlphaMode:
Yeah I noticed that its maven2.lol.bai now or something similar

marcus8448 (approved):

PR #277, opened 2023-07-17
fix: Working Oxygen Bubble Distributor

supersimple33:
This PR is blocked by TeamGalacticraft/MachineLib#9, that is why builds are failing.

I have made the Oxygen Bubble Distributor work. To do this I have:

- Changed the name of `BubbleDistributorScreen` to `OxygenBubbleDistributorScreen`
- I set `bubbleVisible` to private which means that when `setBubbleVisible` is called I can pass that change back onto each client.
- Bubble rendering turns off if the machine redstone is out
- Bubble toggling will now kill the bubbleEntity
- If the `oxygenRequired` to run the machine is less than 1 I now round up so that the machine properly turns off, a weird side effect is that most of the time the machine storage is left with 1 sliver of oxygen
- When the machine cuts out we still keep supplying oxygen until the bubble falls in on itself. (this may be controversial / not intended let me know)
- I added `bubbleVisible` and `size` to the `updateTag` so that those values are persistent and will properly load in when the player joins the world

There are also a number of other small changes to other oxygen systems.
I am not 100% sure about the networking aspect of my impl here in terms of efficiency and multiplayer (I can't test this aspect) so please double check that part.


supersimple33:
I have also made some fixes to the Oxygen Sealer so that it acts correctly. To do this I have added a `tryClearSeal` method to `OxygenSealerBlockEntity` for better tear down when stopping the sealer. Also, I have removed the "Disable Seal" button from the screen, I don't think it makes sense to have it as players should be using the redstone configuration to turn the sealer on/off and this keeps the logic simpler.

PR #287, opened 2023-08-15
fix: Prevent infinite loop on world close caused by wires

### LINKED ISSUE #257: Wires cause infinite loop when closing world
### Mod Loader

Fabric

### Version Information

ac33404a

Reconfirmed in 13285bb

### Log or Crash Report

[TIME][Server thread/WARN] (Galacticraft) Tried to remove wire from removed network!

### Reproduction steps

When using certain wire configurations (typically just a bunch of wires like sometimes a 5x5 square of wires will cause the loop but not every time) the game can end up in an infinite loop where it repeats the error message `Tried to remove wire from removed network!` in the console forever.

benreid24:
Resolves #257 

See [above issue](https://github.com/TeamGalacticraft/Galacticraft/issues/257#issuecomment-1678015422) for problem description.

This PR modifies `WireNetworkImpl` to fetch `Wire` objects from its own map instead of going to the world state. Doing it this way prevents chunks from being (re)loaded while the world is saving and shutting down, which was causing the infinite loop.

Applied the same approach to `PipeNetworkImpl` as pipe networks had the same issue.

### DIFF
--- a/src/main/java/dev/galacticraft/mod/api/pipe/impl/PipeNetworkImpl.java
+++ b/src/main/java/dev/galacticraft/mod/api/pipe/impl/PipeNetworkImpl.java
@@ -48,7 +48,7 @@ import java.util.function.BiFunction;
 public class PipeNetworkImpl implements PipeNetwork {
     private final @NotNull ServerLevel world;
     private final @NotNull Object2ObjectOpenHashMap<BlockPos, Storage<FluidVariant>> insertable = new Object2ObjectOpenHashMap<>();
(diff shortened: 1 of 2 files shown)


benreid24:
Updated the approach to simply bail early instead of rebuilding the networks multiple times from cached state. This approach *may* cause issues if there are legitimate scenarios where a wire can be removed from an unloaded chunk, but I can't think of how that would occur besides on world shutdown.

supersimple33:
These changes work for me

PR #288, opened 2023-08-27
fix: Made rockets no longer stackable

JunePrimavera:


marcus8448 (approved):
Thanks for the patch!

PR #289, opened 2023-12-17
feat: Working Oxygen Bubble Distributor and Sealers

### LINKED ISSUE #270: Game crashes when trying to use bubble distributor and decompressor
### Mod Loader

Fabric

### Version Information

13285bb

### Log or Crash Report

Caused by: java.lang.NullPointerException: Cannot invoke "dev.galacticraft.machinelib.api.storage.slot.SlotGroup.getSlot(int)" because "group" is null

### Reproduction steps

The game will crash when trying to use the bubble distributor or the oxygen decompressor.

supersimple33:
I have made a number of changes so that players can now setup breathing apparatuses on the moon. Fixes #270 if that is still a problem.
- I added Oxygen onto the bubble distributor class name for clarity
- I removed the button from the sealer since that logic is not longer needed and it is covered by using the redstone controls
- Corrected the `OXYGEN_INPUT_SLOT` please confirm the number I switched it to is the correct one that we are trying to use
- Moved syncing logic of the bubble size into `trySyncSize`
- Set the oxygen bubble so that it always requires 1 unit of oxygen to function (certain scales would allow it to function for small radiuses with no o2 in the machine)
- Added logic to bubble distributor to clear oxygen if machine is disabled
- Added clearSeal logic to o2 sealer


supersimple33:
Could someone check that my fix in 7884482 is correct?

PR #305, opened 2024-02-10
fix: License headers

sheazywi:


sheazywi:
succesfully builds and runs with intellij idea

sheazywi:
already done in local commit

PR #310, opened 2024-02-29
Adding the sulfuric acid fluid

marquito73:
I add the sulfuric acid fluid (as it was inside the older Galacticraft mod), with his bucket, this is a list of its properties :

The max length of the fluid is 4 blocks (including source)
The fluid make damage to entities who cross it (1 heart / s)
The fluid interact with entities like water (we can "swim" inside)
The task for this feature : https://github.com/orgs/TeamGalacticraft/projects/2#card-59901681


marcus8448 (approved):
Thanks for the pull request!

Issue #301, opened 2024-01-25
Oil and Fuel liquid doesn't have any effect


sheazywi:
### Mod Loader

Fabric

### Version Information

a555a0e

### Log or Crash Report

Didn't crash

### Reproduction steps

1. Find a natural oil or fuel lake or make a artificial one
2. Walk in it.
3. It's like ur not even walking into a liquid, it doesn't slow u or smt like that. Which is supposed is anormal.

PR #316, opened 2024-05-06
fix: players experienced no viscosity in GC fluids

LeMayday:
Added crude_oil and fuel to water.json. Bug fix for issue #301.


AlphaMode:
I don't really see this as a valid solution. I've been working on a API to solve this. Anyway are fluids aren't water.

LeMayday:
Fair. I wasn't sure exactly how to do this. I think I saw this in one tutorial video for Fabric I watched on YouTube, but I'll admit it wasn't the best. It also seems to make swimming in those liquids like swimming in water, where the texture will be blue despite the fluid texture being green or black or whatever. Hopefully your API will be better.

PR #321, opened 2024-05-26
fix: Space Race Manager screen not rendering as it should

JustAPotota:
- Fix button outlines not rendering
- Fix opening animation being framerate dependant
- Fix text color
- Add (still temporary) team color and team flag menus
- Refactor basically the whole file


src/main/java/dev/galacticraft/mod/client/gui/screen/ingame/ConfirmFlagScreen.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    @Override
+    public void render(GuiGraphics graphics, int mouseX, int mouseY, float delta) {
+        super.render(graphics, mouseX, mouseY, delta);
+        graphics.pose().pushPose();
+        graphics.pose().translate((double) this.width/2 - 24, this.flagTop(), 0.0);
+        //graphics.pose().scale(1.5f, 1.5f, 1.5f);
+        graphics.blit(this.imageLocation, 0, 0, 0, 0, 0, 48, 32, 48, 32);
+        graphics.pose().popPose();
```
marcus8448, line 55: Could the translation be moved into the blit instead of modifying the pose stack?

src/main/java/dev/galacticraft/mod/client/gui/screen/ingame/SpaceRaceScreen.java
```diff
@@ -404,4 +297,125 @@ private enum Menu {
...
+            this.onPress.run();
+        }
+
+        @Override
+        protected void renderWidget(GuiGraphics graphics, int mouseX, int mouseY, float delta) {
+            graphics.pose().pushPose();
+            graphics.pose().translate(this.getX(), this.getY(), 0.0f);
+            graphics.pose().scale(2.0f, 2.0f, 2.0f);
+            graphics.blit(this.imageLocation, 0, 0, 0, 0, 0, 48, 32, 48, 32);
+            graphics.pose().popPose();
```
marcus8448, line 354: Some thing here, but also use the widget's width/height instead of scaling up x2

marcus8448 (changes requested):
Just two small things

src/main/java/dev/galacticraft/mod/client/gui/screen/ingame/SpaceRaceScreen.java
JustAPotota: I tried using the width and height arguments to graphics.blit(), but that just makes it repeat the flag texture to fill space instead of scaling it up. Then that made the X/Y arguments also double, so I changed it to use pose translation instead. This was all taken from how vanilla renders items in the inventory (in GuiGraphics).

src/main/java/dev/galacticraft/mod/client/gui/screen/ingame/ConfirmFlagScreen.java
JustAPotota: See other comment; I was trying out scaling the flag on this screen too, so I used the same method. I can change it back to just passing the position to graphics.blit().

src/main/java/dev/galacticraft/mod/client/gui/screen/ingame/SpaceRaceScreen.java
marcus8448:
```suggestion
            graphics.blit(this.imageLocation, this.getX(), this.getY(), this.width, this.height, 0, 0, 48, 32, 48, 32);
```

marcus8448 (approved):
Thanks!

Issue #324, opened 2024-05-27
rocket does not break when hit in survival mode


LeMayday:
### Mod Loader

Fabric

### Version Information

688f2b2

### Log or Crash Report

N/A

### Reproduction steps

1. Place rocket on launch pad
2. change gamemode to survival
3. attempt to break rocket with fist or sword
4. fail

Issue #323, opened 2024-05-27
Rocket does not break when launchpad is broken


LeMayday:
### Mod Loader

Fabric

### Version Information

688f2b2

### Log or Crash Report

N/A

### Reproduction steps

1. Build 3x3 launch pad
2. Place rocket on launch pad
3. destroy launch pad
4. observe rocket does not break

LeMayday:
this is a tricky one. I can observe a problem in RocketLaunchPadBlockEntity.java. When you put a rocket on the pad, it links just fine via setLinkedRocket, but then when you destroy the pad, it calls hasRocket which doesn't seem to acknowledge that UUID and rocket should no longer be null. Thus the if statement if (pad.hasRocket()) in onRemove() in RocketLaunchPadBlock.java is never entered.

LeMayday:
I'm not familiar with this syntax for instanceof, but my guess is the problem is here in RocketLaunchPadBlock.java:
`if (level.getBlockEntity(center.offset(x, 0, z)) instanceof RocketLaunchPadBlockEntity pad)`

My guess is this constructs a new instance of RocketLaunchPadBlockEntity that does not have the data from before. I will continue looking into this

LeMayday:
nvm just found out this is a thing called pattern matching

LeMayday:
ok actually I wonder if it has anything to do with the fact that it seems to create like 2 new launchpadblock entities every time you place down a launch pad block (so there are 18 in total when you make the multiblock)

LeMayday:
turns out I think the issue was that the call to super() in onRemove destroyed the launchpadblockentity before the rest of the function could figure out whether there was a rocket on it

PR #326, opened 2024-05-28
fix: bugs pertaining to rocket 

LeMayday:
addresses bugs #323 and #324. Also refactors code in RocketLaunchPadBlock.java to improve readability. Also adds a feature where the player will exit the rocket at ground level when dismounting instead of in the air (this was probably the intended behavior)

### DIFF
--- a/src/main/java/dev/galacticraft/mod/content/entity/orbital/RocketEntity.java
+++ b/src/main/java/dev/galacticraft/mod/content/entity/orbital/RocketEntity.java
@@ -291,13 +291,25 @@ public class RocketEntity extends Entity implements Rocket, IgnoreShift, Control
 
     @Override
     public void onBaseDestroyed() {
-        this.spawnAtLocation(GCItems.ROCKET.getDefaultInstance());
+        RocketData data = RocketData.create(this.color(), this.cone(), this.body(), this.fin(), this.booster(), this.engine(), this.upgrade());
+        CompoundTag tag = new CompoundTag();
+        data.toNbt(tag);
+        var rocket = new ItemStack(GCItems.ROCKET);
+        rocket.setTag(tag);
+        this.spawnAtLocation(rocket);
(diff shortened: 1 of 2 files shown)


src/main/java/dev/galacticraft/mod/content/entity/orbital/RocketEntity.java
```diff
@@ -290,12 +291,14 @@ public long getFuel() {
...
     public void onBaseDestroyed() {
-
+        this.spawnAtLocation(GCItems.ROCKET.getDefaultInstance());
+        this.remove(RemovalReason.DISCARDED);
     }
 
     @Override
     public void dropItems(DamageSource damageSource, boolean b) {
-
+        this.spawnAtLocation(GCItems.ROCKET.getDefaultInstance());
```
marcus8448, line 300: (Same thing here)

src/main/java/dev/galacticraft/mod/content/entity/orbital/RocketEntity.java
```diff
@@ -290,12 +291,14 @@ public long getFuel() {
 
     @Override
     public void onBaseDestroyed() {
-
+        this.spawnAtLocation(GCItems.ROCKET.getDefaultInstance());
```
marcus8448, line 294: This spawns a generic rocket item (rather than the type of rocket that was placed).
You'll need to use `::toNbt` to write the rocket part data to the ItemStack.

marcus8448 (changes requested):

src/main/java/dev/galacticraft/mod/content/entity/orbital/RocketEntity.java
LeMayday: Makes sense. I was actually thinking about that earlier today. I can probably address this tomorrow.

marcus8448 (approved):
Thanks!

PR #340, opened 2024-07-13
feat: olivine cave items

JustAPotota:
- Moon Basalt retexture
- Olivine Shard
- Block of Olivine
- Olivine Cluster (drops 4 shards)
- Olivine Basalt and Rich Olivine Basalt (drop 1–2 shards and 3–4 shards)


src/main/generated/assets/galacticraft/models/block/olivine_block.json
```diff
@@ -0,0 +1,7 @@
+{
+  "parent": "minecraft:block/cube_column",
+  "textures": {
+    "end": "galacticraft:block/olivine_block_top",
```
marcus8448, line 4: I'm assuming that this texture isn't ready yet?

marcus8448 (changes requested):
Could you add the blocks/items to the creative mode tab?

marcus8448 (approved):
Thanks!

Issue #333, opened 2024-06-29
Solar panels don't break correctly


supersimple33:
### Mod Loader

Fabric

### Version Information

abfe283

### Log or Crash Report

NA

### Reproduction steps

Place a solar panel down and then a TNT far enough away that it just destroys the base block of the panel. Then the panels will be left behind but in a clear state.

supersimple33:
Probably just need to fix the solar panel part to detect when the `solar_panel_part` isn't connected into a solar panel block anymore.

PR #348, opened 2024-08-24
Made solar panel parts break when the base block breaks

DuckyProgrammer:
Fixed issue https://github.com/TeamGalacticraft/Galacticraft/issues/333

This is my first time contributing to anything large and important like this, so apologies if I did something wrong.

### DIFF
--- a/src/main/java/dev/galacticraft/mod/content/block/machine/SimpleMultiBlockMachineBlock.java
+++ b/src/main/java/dev/galacticraft/mod/content/block/machine/SimpleMultiBlockMachineBlock.java
@@ -24,8 +24,12 @@ package dev.galacticraft.mod.content.block.machine;
 
 import dev.galacticraft.mod.api.block.MultiBlockMachineBlock;
 import dev.galacticraft.mod.api.block.MultiBlockPart;
+import dev.galacticraft.mod.content.GCBlocks;
+import dev.galacticraft.mod.content.block.special.SolarPanelPartBlock;
 import net.minecraft.core.BlockPos;
 import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.level.Explosion;
 import net.minecraft.world.level.Level;
 import net.minecraft.world.level.block.Block;
 import net.minecraft.world.level.block.entity.BlockEntity;
@@ -33,6 +37,7 @@ import net.minecraft.world.level.block.state.BlockState;
 import org.jetbrains.annotations.Unmodifiable;
 
 import java.util.List;
+import java.util.function.BiConsumer;
 
 public class SimpleMultiBlockMachineBlock extends MultiBlockMachineBlock {
     private final List<BlockPos> parts;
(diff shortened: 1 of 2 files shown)


src/main/java/dev/galacticraft/mod/content/block/machine/SimpleMultiBlockMachineBlock.java
```diff
@@ -24,15 +24,20 @@
...
 import net.minecraft.world.level.Level;
 import net.minecraft.world.level.block.Block;
 import net.minecraft.world.level.block.entity.BlockEntity;
 import net.minecraft.world.level.block.state.BlockState;
 import org.jetbrains.annotations.Unmodifiable;
 
 import java.util.List;
+import java.util.function.BiConsumer;
 
 public class SimpleMultiBlockMachineBlock extends MultiBlockMachineBlock {
```
marcus8448, line 42: It would be better to implement this on `MultiBlockMachineBlock` so that all multiblocks can benefit from this change.

src/main/java/dev/galacticraft/mod/content/block/machine/SimpleMultiBlockMachineBlock.java
```diff
@@ -51,6 +56,21 @@ protected SimpleMultiBlockMachineBlock(Properties properties, ResourceLocation f
         this.partState = partBlock.defaultBlockState();
     }
 
+    @Override
+    protected void onExplosionHit(BlockState state, Level world, BlockPos pos, Explosion explosion, BiConsumer<ItemStack, BlockPos> stackMerger) {
```
marcus8448, line 60: It would be better to implement `wasExploded` instead of `onExplosionHit`, as there are explosions that do not break blocks.

src/main/java/dev/galacticraft/mod/content/block/machine/SimpleMultiBlockMachineBlock.java
```diff
@@ -51,6 +56,21 @@ protected SimpleMultiBlockMachineBlock(Properties properties, ResourceLocation f
...
+        if (state.getBlock() != GCBlocks.ADVANCED_SOLAR_PANEL && state.getBlock() != GCBlocks.BASIC_SOLAR_PANEL) {
+            return;
+        }
```
marcus8448, line 63: It can be assumed that the block state is a multiblock (these checks aren't necessary)

src/main/java/dev/galacticraft/mod/content/block/machine/SimpleMultiBlockMachineBlock.java
```diff
@@ -51,6 +56,21 @@ protected SimpleMultiBlockMachineBlock(Properties properties, ResourceLocation f
...
+        for (BlockPos part : parts) {
+            BlockPos actualPartPos = pos.offset(part);
+            if (!(world.getBlockState(actualPartPos).getBlock() instanceof SolarPanelPartBlock)) {
+                return;
```
marcus8448, line 67: Prefer `continue` instead of `return` here, as it's possible that one of that parts was already removed (but others weren't)

src/main/java/dev/galacticraft/mod/content/block/machine/SimpleMultiBlockMachineBlock.java
```diff
@@ -51,6 +56,21 @@ protected SimpleMultiBlockMachineBlock(Properties properties, ResourceLocation f
...
     }
 
+    @Override
+    protected void onExplosionHit(BlockState state, Level world, BlockPos pos, Explosion explosion, BiConsumer<ItemStack, BlockPos> stackMerger) {
+        if (state.getBlock() != GCBlocks.ADVANCED_SOLAR_PANEL && state.getBlock() != GCBlocks.BASIC_SOLAR_PANEL) {
+            return;
+        }
+        for (BlockPos part : parts) {
+            BlockPos actualPartPos = pos.offset(part);
+            if (!(world.getBlockState(actualPartPos).getBlock() instanceof SolarPanelPartBlock)) {
```
marcus8448, line 66: To make this work with all multiblocks you could check if the block entity is a `MultiBlockPart`

marcus8448 (changes requested):
Thanks for the pull request!
Just a few things so that all multiblocks can benefit from this change.

marcus8448 (approved):
Thanks!

PR #355, opened 2024-09-14
fix: made the lander explosion happen on the correct celestial body and disabled it creating fire

DuckyProgrammer:
Fix makes it so when the lander hits too hard the explosion is on the correct celestial body and doesnt create fire


AlphaMode (approved):

megroblog:
top

PR #356, opened 2024-09-17
fix: fix solar panels not showing generation stats when blocked or full

DuckyProgrammer:
Fixed the solar panel screen not showing anything when hovering over the energy bar when the solar panel is blocked or it is full.


DMgaming100:
checked this in game and seems to display the information differently than how other machines do it?

PR #359, opened 2024-09-20
fix: made blocks drop only when the correct tool is used

DuckyProgrammer:
fix: made blocks drop only when the correct tool is used


marcus8448 (approved):
Thanks!

PR #360, opened 2024-09-20
fix: stopped game from crashing when oxygen bubble distributor ran out of oxygen and collapsed

DuckyProgrammer:
fix: stopped game from crashing when oxygen bubble distributor ran out of oxygen and collapsed


marcus8448 (approved):
It would probably be better to avoid setting size to a negative number in the first place, but this works too

PR #361, opened 2024-09-29
Moved the venus sun, moon sun, and earth textures to the environment folder

DMgaming100:
I aint a developer so Im making this a pull request


DMgaming100:
would also appritiate if someone made the solar panels use the gui\celestialbodies textures instead

Issue #392, opened 2025-01-06
Items in firstperson hands turn translucent and eventually disappear when on the Moon


Roelymole:
### Version Information

2228bc687b57fb76b83566b11fe9dba7d76e9e82

### Reproduction steps

Items in the firstperson hands turn translucent and eventually disappear and blocks get darker and darker.
This can be seen in the bottom left and right of the screenshots below, which show the clock go more translucent and then vanish as the time changes.
This may be linked to the footprints which fade and disappear in a similar way.

![2025-01-05_23 50 07](https://github.com/user-attachments/assets/2367b8ea-942c-41f4-803c-07db656497d2)
![2025-01-05_23 50 24](https://github.com/user-attachments/assets/e1908261-0cbb-4ade-8df5-f74a4548d016)
![2025-01-05_23 50 37](https://github.com/user-attachments/assets/8b3e06c0-8459-4745-b25f-4a2675528d76)
![2025-01-05_23 57 43](https://github.com/user-attachments/assets/da815ea5-6782-4297-83e5-b62feae7dd48)

Issue #389, opened 2025-01-05
Incorrectly fall damage just from jumping normally on the Moon


Roelymole:
### Version Information

2228bc687b57fb76b83566b11fe9dba7d76e9e82

### Reproduction steps

Jumps are higher on the Moon due to the lower gravity and this is enough to give fall damage.

The height required before the player takes fall damage should be adjusted so that this doesn't happen.

Jump on the spot and you will take fall damage, even though you fall slower.

Chachavo:
You also hit the ground slower though

PR #394, opened 2025-01-07
fix: bugfixes for #392 and #389

Roelymole:
Bugfixes for #392 and #389.

The issue causing items in the firstperson hands to turn translucent was indeed caused by the footprints - the shader color needed to be reset afterwards.

I have written a mixin to make the fall damage proportional to the gravity which solves the problem for the time being, but may need fine-tuning to account for other factors (i.e. air resistance/atmospheric composition).


Roelymole:
I have moved the commits related to oxygen and suffocation to a separate branch as there ended up being enough of them to warrant their own pull request.

Roelymole:
Prevented passengers of a LanderEntity from being killed by hitting the ground too hard even when landing softly by applying the `#minecraft:fall_damage_immune` tag to the lander.

PR #396, opened 2025-01-10
feat: alpha rocket crafting

JustAPotota:
- Simple rocket recipe type
    - Can handle all the rocket recipes from GC4
    - Serializers, datagen, all that
- Refactor Rocket Workbench
    - Technically kinda supports different recipes, but is currently hardcoded to craft a Tier 1 Rocket

### What still needs to be done:
- Removing modular rockets
- Fixing the Rocket Workbench model
- Full REI integration
    - This PR only adds an exclusion zone
- Multiple rocket recipes (after the alpha)

### Some other ideas:
- Show rocket preview floating on the block itself
- Show tooltip on empty slots, like the Smithing Table does

I probably committed some horrible coding crimes in this one, so I'll do my best to address any comments quickly.


lingyun-create:
So is the rocket able to take off normally now?

Roelymole:
I think it would be better if the rocket preview tab was on the left-hand side, since JEI/REI normally has more content on the right. At the very least, moving the panel down a bit (maybe centered vertically) would probably allow the exclusion zone to only affect the items and not the search bar.

I have a question about the chest slot. Are we not allowing up to 3 chests to be used to craft rockets with additional storage, because this only seems to allow for a single chest?



I love how each component of the rocket is displayed separately in the preview, but it would be even better if the fins could be displayed independently of each other.



The slots for the rocket parts also need to be moved one pixel to the right as they are not aligned with the main inventory slots.

JustAPotota:
> I think it would be better if the rocket preview tab was on the left-hand side, since JEI/REI normally has more content on the right.

That'd make it inconsistent with every other crafting menu going from left to right. There might be enough room to put the preview and output in the main window though.


> I have a question about the chest slot. Are we not allowing up to 3 chests to be used to craft rockets with additional storage, because this only seems to allow for a single chest?

The modular rocket system only supports a rocket having one upgrade, and changing that was out of scope for this PR. Definitely needs to be looked into before the alpha.


> I love how each component of the rocket is displayed separately in the preview, but it would be even better if the fins could be displayed independently of each other.

I think the fins are all one big model? Either way, that would require changing the rocket rendering code, which I don't want to do here. I only threw the preview in since all the pieces were in place for it already. It'll most likely need to be remade or just removed once we figure out what to do with the modular rocket system.


> The slots for the rocket parts also need to be moved one pixel to the right as they are not aligned with the main inventory slots.

Good catch, I'll fix that later tonight :+1:

src/main/java/dev/galacticraft/mod/content/block/special/RocketWorkbench.java
```diff
@@ -62,6 +64,17 @@ protected InteractionResult useWithoutItem(BlockState state, Level level, BlockP
...
+    @Override
+    protected void onRemove(BlockState state, Level level, BlockPos pos, BlockState newState, boolean moved) {
+        if (state.getBlock() != newState.getBlock()) {
+            BlockEntity blockEntity = level.getBlockEntity(pos);
+            if (blockEntity instanceof RocketWorkbenchBlockEntity workbench) {
+                Containers.dropContents(level, pos, workbench.ingredients);
+                Containers.dropContents(level, pos, workbench.chests);
+            }
+        }
+    }
```
marcus8448, line 76: There needs to be a call to `super.onRemove` here (otherwise the block entity is never deleted).

marcus8448 (changes requested):
Sorry this took so long. The patch looks good - just one minor bug.

src/main/java/dev/galacticraft/mod/content/block/special/RocketWorkbench.java
JustAPotota: Fixed :+1:

marcus8448 (approved):
Thanks!

PR #399, opened 2025-01-24
chore: replace hardcoded text with translatable strings, add missing translation providers and correct typos

Roelymole:
- Made hardcoded text translatable:
  - "Creative Only" tooltip for prefueled rocket item
  - "Invalid planet teleport packet received." disconnect message
  - Replaced hardcoded name for the Parachest block entity with the name of the Parachest block.
- Added the %s placeholder to several translatable strings to give more flexibility for translations:
  - "Status: %s"
  - "Traveling to: %s"
  - "Hold \\"%s\\" to slow down!"
  - "Entry Velocity: %s m/s"
  - "%s's Airlock Controller" which was previously hardcoded
- Added translation providers for death messages for deaths caused by Galacticraft damage sources after taking damage from another player or entity e.g. "[Player] ran out of oxygen while trying to escape Evolved Skeleton".
- Changed the generic death message "[Player] succumed to sulfuric acid" to the more interesting and relevant "[Player] was dissolved by sulfuric acid".
- Fixed incorrect lander entry velocity which was half the value it should be since there are 20 ticks per second rather than 10.
- Fixed issues with text on the rocket inventory screen including a more robust method for centering the text.

- Added translation provider for Moon Villagers without a profession.

- Added translation provider for Galacticraft statistics following the vanilla format.

- Corrected typos:
  - RECIPE_VEIWER_DISPLAY_TEXTURE -> RECIPE_VIEWER_DISPLAY_TEXTURE
  - tirer_1_rocket -> tier_1_rocket
  - visiablePredicate -> visiblePredicate


DarianLStephens:
This isn't related to the actual pull here, but is that moon lighting coming from below?
Is it inverted?

Roelymole:
> This isn't related to the actual pull here, but is that moon lighting coming from below? Is it inverted?

I think the Moon has a non-zero ambient light level of 0.1 which is the same as the Nether, so maybe this is what gives you that impression.

PR #400, opened 2025-01-24
feat: prevent elytras from working properly in a vacuum

Roelymole:
This was inspired by the following comment.

> CLtheman1 — 07/01/2025 12:51 PM 
Oh cool, so air resistance is gonna be a thing? That'd be needed for elytras, to make elytras not work in vacuum environments like the moon, space station, and asteroids. Elytras should work even better on Venus due to denser atmosphere.

I have not prevented the player from entering flying mode, but there is no effect on the velocity of the player when there is no atmosphere for the elytra to glide through and firework rockets can't be used either. You can still rotate your body and do all sorts of somersaults, it just doesn't get you any further than a normal jump (which is quite significant on the Moon anyway).


sheazywi:
this would be so cool!

AlphaMode (changes requested):
Instead of using redirects please use [ModifyExpressionValue](https://github.com/LlamaLad7/MixinExtras/wiki/ModifyExpressionValue) instead which has much better compatibility.

Roelymole:
> Instead of using redirects please use [ModifyExpressionValue](https://github.com/LlamaLad7/MixinExtras/wiki/ModifyExpressionValue) instead which has much better compatibility.

I hadn't come across these before, but I will continue to use them where appropriate from now on. Thanks for letting me know!

Issue #364, opened 2024-10-09
Move gui textures to gui folder


DMgaming100:
- [x] Refinery gui is messed up

- [x] Move the `slot/` folder to `gui/slot/`

- [x] Make all the slots in the machines use slot textures from `gui/slot/` instead of being baked into the texture

- [x] Split `solar_panel/default_solar_panels` to 3 textures in `gui/solar_panel/`
 
- [x] Make solar panel use the planet textures from `gui/celestialbodies/` instead of `solar_panel/default_light_sources` & `solar_panel/moon_light_sources`

Roelymole:
Closed by merging #402 and #391.

PR #402, opened 2025-01-26
chore: overhaul GUIs of machines

Roelymole:
- Made all the slots in the machines use slot textures from gui/slot/ instead of being baked into the texture.
  #364

- Adjusted small details such as making the progress indicators more consistent (adjusted positioning as well as progressing in discrete steps).

- Centered the coal slot in the coal generator.

- Flipped the tanks and the slots in the refinery to match the rest of the machines and fixed the positioning of the red markers.

- Removed the shadows from text in the oxygen storage module and used more of the space above for the text (may still need tweaking).

- Rearranged the fuel loader screen so that the title does not overlap with the tank. I would like feedback about the exact positioning.


Roelymole:
### To Do:
- [ ] Change the color of the titles to match vanilla workstations (in MachineLib?)
- [ ] Improve the positioning of several of the titles.
- [ ] Improve the progress indicator of oxygen compressors/decompressors.
- [ ] Improve the way the tabs interact with REI, perhaps by putting them all on the left (see screenshots for examples).
- [ ] Make the style of the oxygen machines more consistent (using the bubble distributor as a basis).
  - [ ] Prevent the status from overflowing for some machines by splitting it across 2 lines.
- [ ] Make the refinery GUI seem less blank by rearranging the positioning and/or including a progress indicator of some kind.
- [ ] Include a "View Recipes for..." button somewhere i.e. the progress arrow, if present, to be consistent with the vanilla workstations.

PR #406, opened 2025-02-01
fix: extinguish fires when there is no oxygen

Roelymole:
- [x] Prevent candles being lit when there is no oxygen.
- [x] Extinguish soul torches/lanterns.
- [x] Extinguish campfires.
- [x] Update the deprecated methods in `TorchBlockMixin`, `LanternBlockMixin` and `CampfireBlockMixin`.
- [x] Extinguish fires which are caused by sources other than fire charges/flint and steel.
- [x] Don't extinguish when there is oxygen provided by an oxygen sealer or oxygen bubble distributor.
- [x] Create a registry system that other mods can use to register which block states should get extinguished and what the extinguished block state is.
- [x] Make furnaces, blast furnaces and smokers extinguishable.


src/main/java/dev/galacticraft/impl/internal/block/ExtinguishableBlockRegistryImpl.java
```diff
@@ -0,0 +1,115 @@
...
+
+        computedEntries = null;
+    }
+
+    @Override
+    public void clear(TagKey<Block> tag) {
+        registeredEntriesTag.remove(tag);
+
+        computedEntries = null;
+    }
```
marcus8448: I don't quite understand what the functional difference between the remove and clear implementations is. `REMOVED` just returns a `null` state, which seems to do the same thing as not having a handler at all?
https://github.com/TeamGalacticraft/Galacticraft/blob/b604a2e1d72f421c7b74ff1360fdc6e3d15f04fc/src/main/java/dev/galacticraft/mod/events/GCEventHandlers.java#L82

src/main/java/dev/galacticraft/impl/internal/block/ExtinguishableBlockRegistryImpl.java
```diff
@@ -0,0 +1,115 @@
...
+import net.minecraft.world.level.block.Block;
+
+import java.util.HashMap;
+import java.util.IdentityHashMap;
+import java.util.Map;
+
+public class ExtinguishableBlockRegistryImpl implements ExtinguishableBlockRegistry {
+    private static final ExtinguishableBlockRegistry.Entry REMOVED = new ExtinguishableBlockRegistry.Entry(state -> null);
+
+    private final Map<Block, ExtinguishableBlockRegistry.Entry> registeredEntriesBlock = new HashMap<>();
```
marcus8448: nit: could also be an `IdentityHashMap` (matching `computedEntries`)

src/main/java/dev/galacticraft/impl/internal/mixin/oxygen/LevelChunkMixin.java
```diff
@@ -113,6 +116,16 @@ private LevelChunkMixin(ChunkPos pos, UpgradeData upgradeData, LevelHeightAccess
...
+        if (blockState.isSolidRender(this.level, pos)) {
+            this.galacticraft$setInverted(pos.getX() & 15, pos.getY(), pos.getZ() & 15, false);
+        }
+    }
+
+    @WrapOperation(method = "setBlockState", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/level/block/state/BlockState;onPlace(Lnet/minecraft/world/level/Level;Lnet/minecraft/core/BlockPos;Lnet/minecraft/world/level/block/state/BlockState;Z)V", ordinal = 0))
+    private void extinguishFire(BlockState newState, Level level, BlockPos pos, BlockState oldState, boolean bl, Operation<Void> original) {
+        if (level.isBreathable(pos) || !GCEventHandlers.extinguishBlock(level, pos, newState)) {
+            original.call(newState, level, pos, oldState, bl);
+        }
```
marcus8448, line 129: This might need some special handling that w.r.t block updates?

If we go from `Level#setBlock` -> `LevelChunk#setBlockState` and then modify the placed state (extinguish), it seems like updates will fire with the _original_ new/placed state.

The vast majority of blocks won't care about this though, so it can be fixed in a later PR.

src/main/java/dev/galacticraft/mod/content/block/entity/machine/CoalGeneratorBlockEntity.java
```diff
@@ -105,16 +114,43 @@ public CoalGeneratorBlockEntity(BlockPos pos, BlockState state) {
...
+            } else {
+                this.setHeat(Math.min(1, this.heat + 0.004));
+            }
+        }
+
+        boolean lit = this.fuelLength > 0;
+        if (this.lit != lit || this.curr != this.prev) {
+            this.lit = lit;
+            this.prev = this.curr;
+            BlockState blockState = this.level.getBlockState(this.worldPosition)
```
marcus8448: Is there a reason this doesn't use the passed `state` parameter?

src/main/java/dev/galacticraft/mod/content/block/entity/machine/CoalGeneratorBlockEntity.java
```diff
@@ -105,16 +114,43 @@ public CoalGeneratorBlockEntity(BlockPos pos, BlockState state) {
...
+                    .setValue(CoalGeneratorBlock.LIT, this.lit)
+                    .setValue(CoalGeneratorBlock.HOT, this.prev);
+            this.level.setBlock(this.worldPosition, blockState, 2);
+        }
+
+        profiler.push("cooling_tick");
```
marcus8448, line 143: This profiler section isn't popped.

src/main/java/dev/galacticraft/mod/content/block/entity/machine/CompressorBlockEntity.java
```diff
@@ -128,10 +139,28 @@ protected void tickConstant(@NotNull ServerLevel world, @NotNull BlockPos pos, @
...
+            BlockState blockState = this.level.getBlockState(this.worldPosition)
+                    .setValue(CompressorBlock.LIT, this.lit);
+            this.level.setBlock(this.worldPosition, blockState, 2);
+        }
+
     }
 
     @Override
     public @NotNull MachineStatus tick(@NotNull ServerLevel level, @NotNull BlockPos pos, @NotNull BlockState state, @NotNull ProfilerFiller profiler) {
+        profiler.popPush("extinguish");
```
marcus8448: This should be a push rather than a pop + push, also never popped.

src/main/java/dev/galacticraft/mod/content/block/entity/machine/CoalGeneratorBlockEntity.java
```diff
@@ -105,16 +114,43 @@ public CoalGeneratorBlockEntity(BlockPos pos, BlockState state) {
...
+        super.tickConstant(level, pos, state, profiler);
+        profiler.push("charge");
+        this.drainPowerToSlot(CHARGE_SLOT);
+        profiler.pop();
+
+        if (this.fuelLength > 0) {
+            if (this.shouldExtinguish(level, pos, state)) {
+                this.fuelLength = 0;
+                this.fuelTime = 0;
+                this.heat = 0;
```
marcus8448: I think heat should be maintained in a vacuum?

src/main/resources/galacticraft.accesswidener
```diff
@@ -17,6 +17,9 @@ accessible field net/minecraft/world/level/levelgen/SurfaceRules$Context biome L
 # final classes
 extendable class net/minecraft/world/SimpleMenuProvider
 
+# private methods
+accessible method net/minecraft/world/level/block/AbstractCandleBlock      getParticleOffsets (Lnet/minecraft/world/level/block/state/BlockState;)Ljava/lang/Iterable;
```
marcus8448: Does it matter that other mods can override `getParticleOffsets` as a protected method? It might be better to implement this as a mixin `@Accessor`

marcus8448 (changes requested):

src/main/java/dev/galacticraft/mod/content/block/entity/machine/CoalGeneratorBlockEntity.java
Roelymole: Probably not, at least not that I can remember.

src/main/java/dev/galacticraft/impl/internal/mixin/oxygen/LevelChunkMixin.java
Roelymole: I'm not quite sure I understand the problem. I also can't exactly remember what I was thinking when I wrote this, but I do feel like some things didn't work if I didn't do it like this.

src/main/resources/galacticraft.accesswidener
Roelymole: Done. I did need to use an `@Invoker` since it is a method, not a field, but the same principle applies.

src/main/java/dev/galacticraft/impl/internal/block/ExtinguishableBlockRegistryImpl.java
Roelymole: `ExtinguishableBlockRegistry` extends the [`net.fabricmc.fabric.api.util.Block2ObjectMap`](https://github.com/FabricMC/fabric/blob/1.21.1/fabric-content-registries-v0/src/main/java/net/fabricmc/fabric/api/util/Block2ObjectMap.java) interface, which has both `remove` and `clear` methods, but I could make one of them an alias for the other if we want them to have the same implementation.

marcus8448: Yeah, I think remove should be made a alias of clear because there's no need to put entries that do nothing into the map.

PR #415, opened 2025-02-09
refactor: auto register BlockItem's and inline item registration

JustAPotota:
- Replace registering `BlockItem`s in `GCItems` with using `BLOCKS.registerWithItem()` in `GCBlocks`
- For other types of items, they're now declared and registered in one go instead of separately
- Remove ~~Herobrine~~ Air Lock Seal item


Roelymole (commented):
This looks much better than the current system to me. May you resolve these few conflicts and then if nobody has any objections could we merge this soon?

JustAPotota:
Should be good to go now :+1:

AlphaMode (approved):

PR #416, opened 2025-02-12
feat: add fluid fog/overlays to oil, fuel and sulfuric acid

Roelymole:
- Added fluid fog/overlays to oil, fuel and sulfuric acid.

- Also made fuel and sulfuric acid fluids translucent.

- Ensured that the fluid fog is not applied when the camera is outside of the fluid e.g. when in third person mode.

- Made it so that you can drown in these fluids using the `#galacticraft:non_breathable` fluid tag, however there is probably still some room for improvement.

- Added a hissing noise and white smoke particles when an entity is in sulfuric acid.


Roelymole:
> Fluid overlays and colour can be set via the `FluidRenderHandler` implementation rather than mixins.

I might be missing something, but as far as I can tell `FluidRenderHandler` only supports the overlay texture when the fluid is viewed through glass/leaves not the overlay texture when the camera is in the fluid and a tint color which changes the color of the fluid itself (note that the fuel is much darker in this screenshot).

src/main/java/dev/galacticraft/mod/mixin/EntityMixin.java
```diff
@@ -137,23 +150,43 @@ public abstract class EntityMixin implements EntityAccessor {
...
+    private void playSulfuricAcidSound() {
+        this.playSound(SoundEvents.LAVA_EXTINGUISH, 0.7F, 1.6F + (level.getRandom().nextFloat() - level.getRandom().nextFloat()) * 0.4F);
+        this.lastAcidSoundPlayTick = this.tickCount;
+    }
+
+    private void addSulfuricAcidParticles() {
+        for (int i = 0; i < 4; i++) {
+            level.addParticle(ParticleTypes.WHITE_SMOKE, true, this.getX() + level.random.nextDouble() - 0.5, Mth.ceil(this.getY()), this.getZ() + level.random.nextDouble() - 0.5, 0.0D, 0.0D, 0.0D);
+        }
+    }
```
marcus8448, line 188: These should be `@Unique`

src/main/java/dev/galacticraft/mod/mixin/client/CameraMixin.java
```diff
@@ -25,22 +25,30 @@
...
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+import com.llamalad7.mixinextras.injector.ModifyExpressionValue;
+import com.llamalad7.mixinextras.sugar.Local;
 
 @Mixin(Camera.class)
 public abstract class CameraMixin {
+    @Shadow private BlockGetter level;
+
+    @Shadow private BlockPos.MutableBlockPos blockPosition;
+
```
marcus8448, line 51: Extraneous shadowed variables?

src/main/java/dev/galacticraft/mod/mixin/client/FogRendererMixin.java
```diff
@@ -24,28 +24,78 @@
...
 
 @Mixin(FogRenderer.class)
 public class FogRendererMixin {
+    @Shadow private static float fogRed;
+    @Shadow private static float fogGreen;
+    @Shadow private static float fogBlue;
 
-    @ModifyArgs(method = "setupColor", at = @At(value = "INVOKE", target = "Lcom/mojang/blaze3d/systems/RenderSystem;clearColor(FFFF)V", remap = false))
-    private static void gc$setupColor(Args args, Camera camera, float partialTicks, ClientLevel clientLevel, int renderDistanceChunks, float bossColorModifier) {
+    @Inject(method = "setupColor", at = @At(value = "INVOKE", target = "Lcom/mojang/blaze3d/systems/RenderSystem;clearColor(FFFF)V", remap = false))
```
marcus8448, line 49: Should this inject sooner than `RenderSystem#clearColor`? Then we can get the `skyDarkness` and night vision modifiers on the fog. (or do we not want those?)

Otherwise the `@At` should probably have an `ordinal=1`.

src/main/java/dev/galacticraft/mod/mixin/client/ScreenEffectRendererMixin.java
```diff
@@ -0,0 +1,98 @@
...
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+@Mixin(ScreenEffectRenderer.class)
+public class ScreenEffectRendererMixin {
+    private static final ResourceLocation OIL_LOCATION = Constant.id("textures/misc/crude_oil_overlay.png");
+    private static final ResourceLocation FUEL_LOCATION = Constant.id("textures/misc/fuel_overlay.png");
+    private static final ResourceLocation ACID_LOCATION = Constant.id("textures/misc/sulfuric_acid_overlay.png");
```
marcus8448, line 60: These should all be `@Unique`

src/main/java/dev/galacticraft/mod/mixin/client/ScreenEffectRendererMixin.java
```diff
@@ -0,0 +1,98 @@
...
+        if (minecraft.player.isEyeInFluid(GCTags.OIL)) {
+            gc$renderFluidOverlay(minecraft, poseStack, OIL_LOCATION);
+        } else if (minecraft.player.isEyeInFluid(GCTags.FUEL)) {
+            gc$renderFluidOverlay(minecraft, poseStack, FUEL_LOCATION);
+        } else if (minecraft.player.isEyeInFluid(GCTags.SULFURIC_ACID)) {
+            gc$renderFluidOverlay(minecraft, poseStack, ACID_LOCATION);
+        }
+    }
+
+    private static void gc$renderFluidOverlay(Minecraft minecraft, PoseStack poseStack, ResourceLocation texture) {
```
marcus8448, line 73: `@Unique`

marcus8448 (changes requested):
Sorry, I thought the API supported camera overlays. Just a couple mixin things.

src/main/java/dev/galacticraft/mod/mixin/client/FogRendererMixin.java
Roelymole: I don't see why we don't want those effects. Do you know if I should move the code that modifies the fog for passengers of a rocket as well?

marcus8448 (commented):
Sorry I didn't catch this the first time around, but we should probably test for the fluids (rather than the tags) for the fog/camera/overlay and sulfuric acid damage so that we don't override other mods' functionality

src/main/java/dev/galacticraft/mod/mixin/client/FogRendererMixin.java
marcus8448: I'm not sure. I think it will look fine either way

Roelymole: It does not look fine... it should be more gradual and less extreme, but you are right about the modifiers not making much of a difference. It might be best to just remove this for now and reimplement it properly later.

https://github.com/user-attachments/assets/2a15d38e-993e-44ca-ac9f-0873bde7c1bc

marcus8448 (approved):
Thanks!

PR #418, opened 2025-02-22
new raw block of resources

DillonPlays84:
adds textures for the blocks of raw 'resource' that are missing


Roelymole:
It would be good if you could familiarise yourself with the [contributing guidelines](https://github.com/TeamGalacticraft/Galacticraft/blob/main/.github/CONTRIBUTING.md), in particular take a look at [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) for how to write your commit messages.

DMgaming100:
sorry Im gonna have to deny those, they seem to be just recolors of each others with just slight tweaks, and dont really follow the style of other raw blocks, as well as not matching the color of the item they are made of

image

PR #423, opened 2025-02-28
feat: implement invalid oxygen setup warning

Roelymole:
- [x] Implemented invalid oxygen setup warning and adjusted lander velocity warning to match.
- [x] Added warning sign symbols - ⚠️ WARNING ⚠️
- [x] Supply the player with oxygen when they are in the lander or a cryogenic chamber.
- [x] Prevented beds from exploding if the atmosphere is non-breathable so that the "What if the oxygen runs out while I'm asleep?" message can actually be seen.
- [x] Allow player to sleep in a bed if there is oxygen provided by a bubble distributor or oxygen sealer - _I'm not sure if we should keep this because it sort of defies the point of the cryogenic chamber._
- [x] Adjust when the oxygen warning is shown.
  - [x] Hide the oxygen warning when in an oxygen bubble/sealed room.
  - [x] Prevent the warning from flashing if the player is not alive.
  - [ ] Don't show the oxygen warning when the player is dead if it was not being displayed before (since the players items are dropped including their oxygen gear making their setup invalid) - _low priority._
- [x] Make a less obtrusive warning for if the player currently has oxygen from an oxygen bubble/sealed room, but has an invalid oxygen setup (some kind of indicator by the oxygen tanks).

https://github.com/user-attachments/assets/60a6b3ce-3ce3-43e8-a3be-8263f7455be0

image


src/main/java/dev/galacticraft/impl/internal/mixin/gear/LivingEntityMixin.java
```diff
@@ -94,6 +98,9 @@ private void galacticraft_oxygenCheck(CallbackInfo ci) {
     @ModifyExpressionValue(method = "baseTick", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/entity/LivingEntity;isEyeInFluid(Lnet/minecraft/tags/TagKey;)Z", ordinal = 0))
     private boolean galacticraft_testForBreathability(boolean original) {
         LivingEntity entity = (LivingEntity) (Object) this;
+        if ((entity.getVehicle() instanceof LanderEntity) || (entity.getInBlockState().getBlock() instanceof CryogenicChamberBlock) || (entity.getInBlockState().getBlock() instanceof CryogenicChamberPart)) {
```
marcus8448, line 101: I think this should check if the player is in cryo sleep (`CryogenicAccessor#isInCryoSleep`) rather than in the block

Roelymole: My thinking was that the player shouldn't suffocate when they wake up inside the cryogenic chamber either, but I can change if you think this doesn't make sense or might be exploited.

marcus8448: That makes sense - if it becomes a problem we can always change it later

marcus8448 (approved):

src/main/java/dev/galacticraft/impl/internal/mixin/gear/LivingEntityMixin.java
marcus8448: That makes sense - if it becomes a problem we can always change it later.

PR #425, opened 2025-03-01
fix: improvements to the rocket

### LINKED ISSUE #419: Rocket spins around after running out of fuel
### Version Information

Latest as of ‎Wednesday, ‎February ‎26, ‎2025, ‏‎15:52:45 UTC
2691cf7269f0940cfe3afdb79616757bc845660c

### Log or Crash Report

_No response_

### Reproduction steps

Reproduction steps:
1. get a rocket
2. give it like 1 bucket of fuel
3. launch
4. beyblade beyblade let it rip

I'm gonna let the videos explain themselves

https://github.com/user-attachments/assets/d7534bc8-daa6-4ecb-a96e-634b6f15b995

https://github.com/user-attachments/assets/d7458b52-aeae-45c9-9056-996dbd029bd9

Roelymole:
- [x] Rewritten the velocity calculations to account for gravity, which resolves #419 (or at least the main problem) as the rockets will no longer float in the air and spin around for very long before exploding.
- [ ] Improve the random rotations when the rocket is in the `LauchStage.FAILED` state.
- [x] Allow the roll of the rocket to be changed with the shift/crouch and space/jump keys.
  - [x] Prevent the camera from moving down when pressing shift in the rocket.
- [x] Adjust the position of the first person camera with the rotation of the rocket.
  - [x] Change the yaw of the camera at the same rate as changing the yaw of the rocket.
  - [x] Invert the controls when in first person mode to be more intuitive.
- [x] Add fire to the explosion when the atmosphere is breathable and use the crash landing damage source for a custom death message.
- [x] Fix issues with the rotations when the rocket is pointing downwards - _I cannot figure out what is causing these issues when the rocket is pointing down, I have tried everything I can think of, so if anyone else can help me see what I am missing that would be great._
  - [x] The player's rotations are out of sync with the rocket when changing the yaw.
  - [x] The rocket's rotation is incorrect if the roll is non-zero.

https://github.com/user-attachments/assets/aaed9572-1ca0-4357-b6ec-c4fd131f6304


Roelymole:
The remaining tasks are not too urgent and I will deal with them eventually, but I have merged this PR now to make the rockets functional again.

PR #439, opened 2025-04-01
Feat: Better Canned Food

maxryan008:
Updates to the canned food stuff


src/main/java/dev/galacticraft/mod/util/TextureUtils.java
```diff
@@ -39,35 +43,84 @@ public static int getAverageColor(ResourceLocation textureLocation) {
...
+        int largest = 0;
+        for (int i = 1; i < k; i++) {
+            if (centroids.get(i).length() > centroids.get(largest).length()) {
+                largest = i;
+            }
+        }
+        return centroids.get(largest);
+    }
+
+    private static float distanceSquared(Vector3f a, Vector3f b) {
```
Roelymole, line 108: I think you can get rid of this method and use [this method](https://joml-ci.github.io/JOML/apidocs/org/joml/Vector3f.html#distanceSquared(org.joml.Vector3fc)) instead.

e.g. `a.distanceSquared(b)` where `a` and `b` are both `Vector3f`s.

PR #462, opened 2025-04-26
feat: prevent oxygen tank drain when wearing infinite capacity tank

DuckyProgrammer:
If you have a non-infinite oxygen tank in slot 1 and infinite tank in slot 2, it used to drain the non-infinite tank until it was empty then use the infinite one.

### DIFF
--- a/src/main/java/dev/galacticraft/impl/internal/mixin/gear/LivingEntityMixin.java
+++ b/src/main/java/dev/galacticraft/impl/internal/mixin/gear/LivingEntityMixin.java
@@ -33,14 +33,13 @@ import dev.galacticraft.mod.content.block.special.CryogenicChamberBlock;
 import dev.galacticraft.mod.content.block.special.CryogenicChamberPart;
 import dev.galacticraft.mod.content.entity.damage.GCDamageTypes;
 import dev.galacticraft.mod.content.entity.orbital.lander.LanderEntity;
-import dev.galacticraft.mod.content.item.OxygenTankItem;
+import dev.galacticraft.mod.content.item.InfiniteOxygenTankItem;
 import dev.galacticraft.mod.tag.GCTags;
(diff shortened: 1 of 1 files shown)


src/main/java/dev/galacticraft/impl/internal/mixin/gear/LivingEntityMixin.java
```diff
@@ -135,6 +137,17 @@ private void galacticraft_modifyAirLevel(int air, CallbackInfoReturnable<Integer
 
         if (this.galacticraft$hasMaskAndGear()) {
             InventoryStorage tankInv = InventoryStorage.of(galacticraft$getOxygenTanks(), null);
+            for (int i = 0; i < tankInv.getSlotCount(); i++) {
+                ItemStack stack = tankInv.getSlot(i).getResource().toStack();
+                if (stack.getItem() instanceof OxygenTankItem) {
+                    StorageView<FluidVariant> view = OxygenTankItem.getStorage(stack);
+                    if (view.getCapacity() == Long.MAX_VALUE) {
```
Roelymole, line 144: If you just check if it is an instance of `InfiniteOxygenTank` then you don't need the storage view code, making things a bit simpler.

src/main/java/dev/galacticraft/impl/internal/mixin/gear/LivingEntityMixin.java
```diff
@@ -135,6 +137,17 @@ private void galacticraft_modifyAirLevel(int air, CallbackInfoReturnable<Integer
 
         if (this.galacticraft$hasMaskAndGear()) {
             InventoryStorage tankInv = InventoryStorage.of(galacticraft$getOxygenTanks(), null);
+            for (int i = 0; i < tankInv.getSlotCount(); i++) {
```
Roelymole, line 140: This loop could start from `int i = 1`

Roelymole (approved):

PR #464, opened 2025-04-29
fix: adjust conflicting pirate speak translations

Roelymole:
Adjust the pirate speak translations that conflict with things in Galacticraft:
- The pirate speak translation for coal is "Fuel", which could get confused with the liquid fuel as mentioned in #230
- The word iron gets replaced with "steel" in the pirate speak translation, which also causes confusion between compressed iron and compressed steel, for example.

Including the overriding translations in the file `resources/assets/minecraft/lang/en_pt.json` seems like the most elegant solution, because this probably isn't something that should go through Crowdin.

I have changed "Fuel" to "Lump o' coal" rather than just "Coal", because that doesn't seem piratical enough, but other than that the translations are just the vanilla translations with a single replacement.


Roelymole:
> I think it would be better for Galacticraft pirate speak to rename (real) steel/fuel to something else, as other mods may name their iron "steel", etc. However, I must admit that I have no clue what to rename them to.

Fair enough, I clearly hadn't thought this through for very long.

> As for the actual patch - if we do go this way, we'll need some extra build logic for when crowdin translations are actually pulled - this en_pt.json will conflict with the crowdin one at build-time (when `src/main/resources` and `src/generated` are merged in the jar)

It wouldn't have been a problem anyway, right? Since this is in the `assets/minecraft/lang` folder and the other translations would be in the `assets/galacticraft/lang` folder.

marcus8448:
> It wouldn't have been a problem anyway, right? Since this is in the `assets/minecraft/lang` folder and the other translations would be in the `assets/galacticraft/lang` folder.

Oops - yeah, it would have been fine. I didn't notice the directory

Issue #457, opened 2025-04-25
Death when staying in planetary selection menu


DuckyProgrammer:
### Version Information

0e2187f

### Reproduction steps

Go into survival mode
Place a rocket
Fuel it (or use the creative rocket)
Launch
Stay in the menu
Eventually you die from fall damage

PR #466, opened 2025-05-05
fix: Death when staying in planetary selection menu #457

CoderFroty:


Roelymole (commented):
If anything this would probably be better in the `tick` method of `CelestialScreen` [here](https://github.com/TeamGalacticraft/Galacticraft/blob/2cf109e7ab1292c66cda9bee589765c496ffe309/src/main/java/dev/galacticraft/mod/client/gui/screen/ingame/CelestialScreen.java#L252), to save checking the screen each tick.

We might want to do something more like the `changeDimension` method of `ServerPlayer` eventually.

src/main/java/dev/galacticraft/mod/client/gui/screen/ingame/CelestialScreen.java
```diff
@@ -253,6 +253,11 @@ public void tick() {
         this.translationX = 0.0F;
         this.translationY = 0.0F;
 
+        if (Minecraft.getInstance().screen instanceof CelestialSelectionScreen) {
```
Roelymole, line 256: Since `CelestialSelectionScreen` extends `CelestialScreen` you don't need this if statement.

src/main/java/dev/galacticraft/mod/client/gui/screen/ingame/CelestialScreen.java
```diff
@@ -252,7 +252,8 @@ protected void unselectCelestialBody() {
     public void tick() {
         this.translationX = 0.0F;
         this.translationY = 0.0F;
-
+        assert Minecraft.getInstance().player != null;
+        Minecraft.getInstance().player.setDeltaMovement(new Vec3(0.0D, 0.0D, 0.0D));
```
Roelymole, line 256: Sorry I didn't realise/mention this earlier, but you can also use `this.minecraft` rather than `Minecraft.getInstance()` since `CelestialScreen` extends `Screen`.

Also, instead of asserting `player != null` there should be an if statement, because assert will not prevent a NullPointerError.

Roelymole (approved):
Thanks for making those changes.

src/main/java/dev/galacticraft/mod/client/gui/screen/ingame/CelestialScreen.java
```diff
@@ -252,7 +253,9 @@ protected void unselectCelestialBody() {
     public void tick() {
         this.translationX = 0.0F;
         this.translationY = 0.0F;
-
+        if (this.minecraft.player != null && this.minecraft.player.getY() >= ESCAPE_HEIGHT) {
```
Roelymole, line 256: This latest change seems to have had the opposite effect of making the player die sooner. I think changing `ESCAPE_HEIGHT` to a smaller value such as `REENTRY_HEIGHT` should prevent this with a very generous margin for error.

src/main/java/dev/galacticraft/mod/client/gui/screen/ingame/CelestialScreen.java
```diff
@@ -65,6 +65,7 @@
 
 import static dev.galacticraft.mod.Constant.CelestialScreen.*;
 import static dev.galacticraft.mod.Constant.ESCAPE_HEIGHT;
```
Roelymole: You can remove this import since it is now unused.

Roelymole (approved):
Let me know when you have finished making changes and are ready for this to be merged.

PR #468, opened 2025-05-09
feat: oxygen gear for pets

Roelymole:
### Tasks - possibly for separate pull requests:
- [x] Reduce the oxygen level of parrots when they are on the player's shoulders.
- [x] Adjust oxygen consumption for pets.
  - [x] Lower consumption rates.
  - [x] Don't consume oxygen when the owner is not nearby?
- [x] Prevent pets from being fed when wearing an oxygen mask or in a non-breathable area (except from a can).
- [x] Show oxygen levels of other players and pets in WTHIT/WAILA tooltip.
- [ ] Allow pets to travel in a rocket with their owner - _temporary, would probably get reworked._
  - [ ] Either one dog or one cat.
  - [x] Up to two parrots on the player's shoulders (mostly functional to begin with).
- [ ] Fix issues synchronising the gear inventory when there are no updates, for example, when the entity is in a breathable area or is wearing an infinite oxygen tank.


src/main/java/dev/galacticraft/mod/content/GCAccessorySlots.java
```diff
@@ -32,18 +32,19 @@
...
-    public static final int OXYGEN_TANK_2_SLOT = 7;
+    public static final int OXYGEN_MASK_SLOT = 0;
+    public static final int OXYGEN_GEAR_SLOT = 1;
+    public static final int OXYGEN_TANK_1_SLOT = 2;
+    public static final int OXYGEN_TANK_2_SLOT = 3;
+    public static final int ACCESSORY_SLOT_START = 4;
+    public static final int ACCESSORY_SLOT_END = 7;
+    public static final int THERMAL_ARMOR_SLOT_START = 8;
+    public static final int THERMAL_ARMOR_SLOT_END = 11;
+    public static final int PET_THERMAL_SLOT = 3;
```
marcus8448, line 43: Could you use these constants when creating the gear inventories in the *Player mixins?
Otherwise I fear these will get out of sync with the actual inventory

src/main/java/dev/galacticraft/impl/internal/mixin/gear/LivingEntityMixin.java
```diff
@@ -201,6 +272,63 @@ private void galacticraft_canFreezeThermalPadding(CallbackInfoReturnable<Boolean
...
+
+            if (tankInv.getContainerSize() > 1) {
+                ItemStack itemStack = tankInv.getItem(1);
+                String tankSize = "";
+                if (itemStack.is(GCItemTags.OXYGEN_TANKS)) {
+                    tankSize = itemStack.getDescriptionId().replace("item.galacticraft.", "");
+                }
+                this.entityData.set(DATA_TANK_2_SIZE_ID, tankSize);
+            }
+        }
```
marcus8448, line 311: This is not blocking (more of a later issue), but we probably don't want to use a item tag to recognize oxygen tanks in the future.
It would be better to have all oxygen tanks implement an interface to provide sizing/rendering data.

src/main/java/dev/galacticraft/impl/internal/mixin/gear/TamableAnimalMixin.java
```diff
@@ -0,0 +1,100 @@
...
+    @Override
+    public Container galacticraft$getAccessories() {
+        return this.accessoryInv;
+    }
+
+    @Override
+    public long galacticraft$oxygenConsumptionRate() {
+        TamableAnimal animal = (TamableAnimal) (Object) this;
+        LivingEntity owner = animal.getOwner();
+        if (owner != null && !owner.isSpectator() && animal.distanceToSqr(owner) >= 4096.0) {
```
marcus8448: I assume we wouldn't want animals to die while players are disconnected?
```suggestion
        if (owner == null || (!owner.isSpectator() && animal.distanceToSqr(owner) >= 4096.0)) {
```

Roelymole: Definitely not, thanks!

marcus8448 (approved):

src/main/java/dev/galacticraft/impl/internal/mixin/gear/LivingEntityMixin.java
Roelymole: I'll work on this in a separate PR.

Issue #461, opened 2025-04-26
Oxygen tanks deplete and Invalid Oxygen Setup warning appears when in sealed area


DuckyProgrammer:
### Version Information

c51c7bc

### Reproduction steps

Go to an environment where there is no oxygen
Make a small shack and place an oxygen sealer with an infinite tank and infinite battery
Observe invalid setup warning
Put on oxygen setup with any tank (non infinite)
Hover over oxygen bar, press shift, observe the value dropping

DuckyProgrammer:
Looking more into this, I think its client side. When I relog the oxygen tank goes back to its previous capacity.

DuckyProgrammer:
Working on fixing this

Roelymole:
This issue might be closely linked to #433.

davo899:
I've looked through the code and I can't find anywhere where it actually sends the information to the client about blocks that are breathable due to oxygen sealers.

Roelymole:
It might have been when #320 was merged.

davo899:
Working on fixing this.

Issue #433, opened 2025-03-20
Placing and breaking a block inside an oxygen bubble makes it non-breathable


Roelymole:
### Version Information

9a06a62610df1940b70485644742fb7dea5340e1

### Reproduction steps

1. Place down an oxygen bubble distributor and supply it with electricity and oxygen.
2. Place a block inside the oxygen bubble and break it again.
3. Put your head in the space that the block used to occupy and observe that the oxygen warning is shown.

Additionally, placing non-full blocks in a space which has oxygen will make that location non-breathable.

![Image](https://github.com/user-attachments/assets/b1e3cfc5-008c-42ff-a7d5-bf445c3f511f)

Leaving and rejoining the world causes the oxygen to be recomputed.

Roelymole:
Good spot. That definitely doesn't look right.

PR #482, opened 2025-06-09
fix: double iteration of loop counters in galacticraft$syncOxygenPacketsToClient

davo899:


davo899:
Fixes #433

Roelymole:
Does this also fix #461?

PR #483, opened 2025-06-12
fix: oxygen sealer mechanics

### LINKED ISSUE #461: Oxygen tanks deplete and Invalid Oxygen Setup warning appears when in sealed area

davo899:
Resolves #461

### DIFF
--- a/src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenSealerBlockEntity.java
+++ b/src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenSealerBlockEntity.java
@@ -101,7 +101,6 @@ public class OxygenSealerBlockEntity extends MachineBlockEntity {
     @Override
     public void setLevel(Level level) {
         super.setLevel(level);
-        this.sealCheckTimer = SEAL_CHECK_TIME;
         if (!level.isClientSide) {
             level.galacticraft$getSealerManager().addSealer(this);
         }
@@ -137,11 +136,6 @@ public class OxygenSealerBlockEntity extends MachineBlockEntity {
             return GCMachineStatuses.BLOCKED;
         }
         this.blocked = false;
-
-        // Timer for display on UI
-        if (this.sealCheckTimer-- <= 0) {
-            this.sealCheckTimer = SEAL_CHECK_TIME;
-        }
 
         if (this.hasEnergy) {
             this.consumeEnergy();
@@ -204,6 +198,7 @@ public class OxygenSealerBlockEntity extends MachineBlockEntity {
     }
 
     public int getSealTickTime() {
-        return this.sealCheckTimer;
+        if (level == null) return 0;
+        return (int) (level.getGameTime() % SEAL_CHECK_TIME);
     }
 }
--- a/src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenSealerBlockEntity.java
+++ b/src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenSealerBlockEntity.java
@@ -61,7 +61,6 @@ public class OxygenSealerBlockEntity extends MachineBlockEntity {
     public static final long MAX_OXYGEN = FluidUtil.bucketsToDroplets(20);
     public static final int SEAL_CHECK_TIME = 20;
 
-    private int sealCheckTimer = SEAL_CHECK_TIME;
     private boolean isSealed = false;
     private boolean hasEnergy = false;
     private boolean hasOxygen = false;
@@ -190,7 +189,7 @@ public class OxygenSealerBlockEntity extends MachineBlockEntity {
     }
 
     public boolean hasOxygen() {
-        return this.hasEnergy;
+        return this.hasOxygen;
     }
 
     public boolean isBlocked() {
--- a/src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenSealerBlockEntity.java
+++ b/src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenSealerBlockEntity.java
@@ -120,7 +120,6 @@ public class OxygenSealerBlockEntity extends MachineBlockEntity {
     protected @NotNull MachineStatus tick(@NotNull ServerLevel level, @NotNull BlockPos pos, @NotNull BlockState state, @NotNull ProfilerFiller profiler) {
         // Check if the machine has enough energy
         if (!this.energyStorage().canExtract(Galacticraft.CONFIG.oxygenSealerEnergyConsumptionRate())) {
-            //recalculate area to avoid issues with fake oxygen
             this.hasEnergy = false;
             return MachineStatuses.NOT_ENOUGH_ENERGY;
         }
@@ -128,14 +127,12 @@ public class OxygenSealerBlockEntity extends MachineBlockEntity {
 
         // Check if the oxygen tank is empty
         if (this.fluidStorage().slot(OXYGEN_TANK).isEmpty()) {
-            //recalculate area to avoid issues with fake oxygen
             this.hasOxygen = false;
             return GCMachineStatuses.NOT_ENOUGH_OXYGEN;
         }
         this.hasOxygen = true;
 
         if (!level.getBlockState(pos.offset(0, 1, 0)).isAir()) {
-            //recalculate area to avoid issues with fake oxygen
             this.blocked = true;
             return GCMachineStatuses.BLOCKED;
         }
--- a/src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenSealerBlockEntity.java
+++ b/src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenSealerBlockEntity.java
@@ -121,9 +121,7 @@ public class OxygenSealerBlockEntity extends MachineBlockEntity {
         // Check if the machine has enough energy
         if (!this.energyStorage().canExtract(Galacticraft.CONFIG.oxygenSealerEnergyConsumptionRate())) {
             //recalculate area to avoid issues with fake oxygen
-            if (this.hasEnergy) {
-                this.hasEnergy = false;
-            }
+            this.hasEnergy = false;
             return MachineStatuses.NOT_ENOUGH_ENERGY;
         }
         this.hasEnergy = true;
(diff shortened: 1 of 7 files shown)


davo899:
Fixes #461

Roelymole:
If you use closes or resolves instead of fixes, the issue will be linked properly so that merging the pull request closes the issue. The same applies if you write it in a commit message.

src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenSealerBlockEntity.java
```diff
@@ -128,12 +119,10 @@ protected void tickConstant(@NotNull ServerLevel level, @NotNull BlockPos pos, @
     @Override
     protected @NotNull MachineStatus tick(@NotNull ServerLevel level, @NotNull BlockPos pos, @NotNull BlockState state, @NotNull ProfilerFiller profiler) {
         // Check if the machine has enough energy
-        SealerManager manager = ((GCLevelAccessor) level).getSealerManager();
         if (!this.energyStorage().canExtract(Galacticraft.CONFIG.oxygenSealerEnergyConsumptionRate())) {
             //recalculate area to avoid issues with fake oxygen
```
Roelymole: This comment seems outdated and should probably be removed.

src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenSealerBlockEntity.java
```diff
@@ -128,12 +119,10 @@ protected void tickConstant(@NotNull ServerLevel level, @NotNull BlockPos pos, @
...
             if (this.hasEnergy) {
```
Roelymole: This if statement is probably unnecessary now.

src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenSealerBlockEntity.java
```diff
@@ -144,7 +133,6 @@ protected void tickConstant(@NotNull ServerLevel level, @NotNull BlockPos pos, @
             //recalculate area to avoid issues with fake oxygen
             if (this.hasOxygen) {
```
Roelymole: The same thing applies here, as above.

src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenSealerBlockEntity.java
```diff
@@ -154,27 +142,23 @@ protected void tickConstant(@NotNull ServerLevel level, @NotNull BlockPos pos, @
             //recalculate area to avoid issues with fake oxygen
             if (!this.blocked) {
```
Roelymole: Here too.

src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenSealerBlockEntity.java
```diff
@@ -154,27 +142,23 @@ protected void tickConstant(@NotNull ServerLevel level, @NotNull BlockPos pos, @
...
             return GCMachineStatuses.BLOCKED;
         }
         this.blocked = false;
 
-        // Update sealing status periodically
+        // Timer for display on UI
         if (this.sealCheckTimer-- <= 0) {
             this.sealCheckTimer = SEAL_CHECK_TIME;
-            manager.recalculateSealingStatus(pos, level);
         }
```
Roelymole: Can you ensure this `sealCheckTimer` is synced with the `SealerManager`'s `sealCheckTimer`?

If not it might just be better to use the `SealerManager`'s in the UI directly.

src/main/java/dev/galacticraft/mod/mixin/LevelMixin.java
```diff
@@ -31,10 +31,10 @@
 @Mixin(Level.class)
 public class LevelMixin implements GCLevelAccessor {
     @Unique
-    private final SealerManager sealerManager = new SealerManager();
+    private final SealerManager sealerManager = new SealerManager((Level)(Object)this);
```
Roelymole: It might be slightly neater to separate the casts with spaces, like so: `(Level) (Object) this`

src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenSealerBlockEntity.java
davo899: The issue is client vs server side, sealer manager shouldn't really exist on the client side at all but im not entirely sure how to rework that properly. Any thoughts?

src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenSealerBlockEntity.java
davo899: Im a bit confused about this, aren't these if statements controlling the status in the sealer's UI?

davo899: Nevermind, I get what you mean now lol

src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenSealerBlockEntity.java
Roelymole: Good point. Perhaps the simplest thing to do would be to recheck whenever the `level.getGameTime() % SEAL_CHECK_TIME == 0`.

Roelymole:
Will you have a chance to look at my other comments at some point soon?

davo899:
I will tomorrow, sorry

src/main/java/dev/galacticraft/mod/mixin/LevelMixin.java
```diff
@@ -31,10 +31,10 @@
 @Mixin(Level.class)
 public class LevelMixin implements GCLevelAccessor {
     @Unique
-    private final SealerManager sealerManager = new SealerManager();
+    private final SealerManager sealerManager = new SealerManager((Level) (Object) this);
```
Roelymole, line 34: Also, for mixins you should use the `@Final` decorator from `org.spongepowered.asm.mixin.Final` rather than the final keyword.

src/main/java/dev/galacticraft/mod/machine/SealerManager.java
```diff
@@ -23,147 +23,128 @@
...
-        for (Map.Entry<BlockPos, OxygenSealerBlockEntity> entry : this.sealers.entrySet()) {
-            if (entry.getKey().distSqr(pos) <= OxygenSealerBlockEntity.SEALER_RANGE * OxygenSealerBlockEntity.SEALER_RANGE) {
-                return entry.getKey();
-            }
+    private static final double MAX_SEALER_VOLUME = 1024;
+
+    public void tick() {
+        // Update sealing status periodically
+        if (this.sealCheckTimer-- <= 0) {
+            this.sealCheckTimer = SEAL_CHECK_TIME;
```
Roelymole, line 68: This is still not necessarily in sync, but if you were to check whether `level.getGameTime() % SEAL_CHECK_TIME == 0` here it should work.

Setting `SEAL_CHECK_TIME` to a larger value for testing should make it more obvious if they are in sync or not.

Roelymole, line 68: This is still not necessarily in sync, but if you were to check whether `level.getGameTime() % SEAL_CHECK_TIME == 0` here it should work.

Setting `SEAL_CHECK_TIME` to a larger value for testing should make it more obvious if they are in sync or not.

src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenSealerBlockEntity.java
```diff
@@ -222,14 +189,15 @@ public boolean hasEnergy() {
...
     }
 
     public boolean isBlocked() {
         return this.blocked;
     }
 
     public int getSealTickTime() {
-        return this.sealCheckTimer;
+        if (level == null) return 0;
+        return (int) (level.getGameTime() % SEAL_CHECK_TIME);
```
Roelymole, line 201: Thinking about it, we want to count down to the next check, so it should be `SEAL_CHECK_TIME - (int) (level.getGameTime() % SEAL_CHECK_TIME)`.

src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenSealerBlockEntity.java
```diff
@@ -128,53 +117,32 @@ protected void tickConstant(@NotNull ServerLevel level, @NotNull BlockPos pos, @
...
 
         // Check if the oxygen tank is empty
         if (this.fluidStorage().slot(OXYGEN_TANK).isEmpty()) {
-            //recalculate area to avoid issues with fake oxygen
-            if (this.hasOxygen) {
-                this.hasOxygen = false;
-                manager.recalculateSealingStatus(pos, level);
-            }
+            this.hasOxygen = false;
             return GCMachineStatuses.NOT_ENOUGH_OXYGEN;
```
Roelymole, line 129: This is not currently taking into account how much oxygen there is, just that there is some, meaning that the sealer could run on less oxygen than it technically requires.

At the moment the `consumeOxygen` method extracts the amount returned by `Galacticraft.CONFIG.oxygenSealerOxygenConsumptionRate()`, so at the very least you should check if you can extract that much (similarly to the energy above).

I imagine in the future we will want to scale the amount of oxygen required by how many blocks are being sealed which will probably require a slightly different approach, but if you have time maybe you could try doing this now. This amount would have to be shared between sealers if there are multiple sealers in the same space, which might be slightly fiddly.

src/main/java/dev/galacticraft/mod/machine/SealerManager.java
```diff
@@ -23,147 +23,128 @@
...
+            if (!sealer.hasEnergy()) continue;
+            if (!sealer.hasOxygen()) continue;
+            if (sealer.isBlocked()) continue;
+
+            // Flood fill to find all blocks this sealer is trying to seal
+            SpaceToSeal spaceToSeal = new SpaceToSeal(sealer);
+            while (!spaceToSeal.floodFillQueue.isEmpty()) {
+                BlockPos pos = spaceToSeal.floodFillQueue.pollFirst();
+                if (spaceToSeal.blocksToSeal.contains(pos)) continue;
+                if (!level.getBlockState(pos).isAir()) continue;
```
Roelymole, line 100: The only blocks for which the `isAir()` method returns true are `Blocks.AIR`, `Blocks.VOID_AIR` and `Blocks.CAVE_AIR`, which means that a room could be sealed with something like a torch as part of the wall.

src/main/java/dev/galacticraft/mod/mixin/LevelMixin.java
marcus8448: I don't think `@Final` applies here - its not a shadowed field.

PR #486, opened 2025-06-27
fix: update gradle and deps

JustAPotota:
Fixes the project not loading. Also adds license headers so GitHub will stop emailing me about failed PR checks.


marcus8448 (approved):
Thanks!

PR #488, opened 2025-06-27
Fixed all mavens

maxryan008:
All mavens have been updated to their latest versions and the new maven url is https://maven.teamgalacticraft.org


SebaSphere:
mb guys, forgor to renew the secondary domain so it was down

PR #495, opened 2025-07-05
feat: improved Earth in satellite sky and when leaving the overworld

Roelymole:
https://github.com/user-attachments/assets/a7a9e435-274a-492d-a7fc-2b2941df5b81

- Greatly increased the size of the Earth to a more realistic size.
- Added an atmosphere around the edge of the Earth.
- Separated the clouds from the ground in a slowly moving layer.

## To Do:
- [ ] Prevent the edges of the Earth from getting cut off for low render distances.

### Improve the transition when leaving the overworld
- [ ] Improve the sky color transition.
- [ ] Render stars correctly i.e. not visible/barely visible when the sky is light.

- [ ] Allow for resource packs to add circular planets using a mask layer.


Roelymole:
If you are using a development environment you can use the command `git fetch origin pull/ID/head:name_of_your_choice` using the ID of the pull request, which in this case is 495 and then `git checkout name_of_your_choice`.

From here you would run the game like normal with the `.\gradlew runClient`, or `.\gradlew.bat runClient` command for Windows.

I have uploaded the latest build of this PR to [Dropbox](https://www.dropbox.com/scl/fi/55jsj6yl9ympb8ki5jlwc/Galacticraft-5.0.0-prealpha-e1ff437e.jar?rlkey=v8ttlj0u5lplxgfvsdxrl340t&st=1bs4vwdl&dl=0), but I would recommend setting up a development environment by cloning the repository locally if you find you want to test things out often.

PR #498, opened 2025-07-23
feat: omnishape compat to add framable pipes/wires

maxryan008:
Added compat between Omnishape and Galacticraft 5.


build.gradle.kts
```diff
@@ -249,6 +258,8 @@ dependencies {
 
     modCompileOnly("mezz.jei:jei-$minecraftVersion-fabric-api:$jeiVersion")
 
+    modCompileOnly("dev.maximus:omnishape:1.2.4")
```
marcus8448, line 261: This could use a property/placeholder for the version (like the other dependencies).

src/main/java/dev/galacticraft/mod/GalacticraftMixinPlugin.java
```diff
@@ -41,6 +44,9 @@ public String getRefMapperConfig() {
 
     @Override
     public boolean shouldApplyMixin(String targetClassName, String mixinClassName) {
+        if ((mixinClassName.contains("ExplosionDamageCalculatorMixin") || mixinClassName.contains("ServerPlayerGameModeMixin")) && !OMNISHAPE_LOADED) {
+            return false;
+        }
```
marcus8448, line 49: A bit nitpicky but I think this is clearer.
```suggestion
        if ((mixinClassName.contains("ExplosionDamageCalculatorMixin") || mixinClassName.contains("ServerPlayerGameModeMixin"))) {
            return OMNISHAPE_LOADED;
        }
```
It would probably be a good idea to include the `omnishape.` package prefix in the test (since the mixins are named after their targets)

src/main/java/dev/galacticraft/mod/api/block/PipeShapedBlock.java
```diff
@@ -109,10 +131,21 @@ public void neighborChanged(BlockState state, Level level, BlockPos pos, Block b
 
     @Override
     protected @NotNull VoxelShape getShape(BlockState state, BlockGetter world, BlockPos pos, CollisionContext context) {
-        if (world.getBlockEntity(pos) instanceof Connected connected) {
-            return this.shapes[generateAABBIndex(connected)];
+        VoxelShape baseShape = Shapes.empty();
```
marcus8448, line 134: Although this theoretically shouldn't ever occur, it's probably better to fallback to `this.shapes[0]` (as before) rather than empty.

src/main/java/dev/galacticraft/mod/mixin/client/MultiPlayerGameModeMixin.java
```diff
@@ -0,0 +1,71 @@
...
+import net.minecraft.world.phys.shapes.VoxelShape;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+
+import static dev.galacticraft.mod.api.block.PipeShapedBlock.preciseHit;
+
+@Mixin(MultiPlayerGameMode.class)
+public abstract class MultiPlayerGameModeMixin {
```
marcus8448, line 30: This should probably be filtered in the mixin plugin as well

src/main/java/dev/galacticraft/mod/content/block/entity/networked/WireBlockEntity.java
```diff
@@ -48,6 +53,25 @@ public class WireBlockEntity extends BlockEntity implements Wire, EnergyStorage
...
+    }
+
+    public @Nullable Boolean getAndClearTargetingFrame() {
+        Boolean result = this.wasTargetingFrame;
+        this.wasTargetingFrame = null;
+        return result;
+    }
+
+    @Nullable
+    private OmnishapeData frameOverlay = null;
```
marcus8448, line 73: If Omnishape isn't loaded `OmnishapeData` won't exist (I'm pretty sure a field will always cause a class load but idk)

marcus8448 (changes requested):
A couple things

src/main/java/dev/galacticraft/mod/content/block/entity/networked/WireBlockEntity.java
maxryan008: hmm maybe it was just due to the debug environement but that never threw an issue when i removed the mod from my dev environment. i wasnt 100% sure how to do it without that but after your recent suggestion of making the omnishape api have its own system for blocks like that then yes it should be possible to remove that.

src/main/java/dev/galacticraft/mod/content/block/entity/networked/WireBlockEntity.java
```diff
@@ -177,4 +208,40 @@ public long getAmount() {
...
+
+    public boolean hasOverlay() {
+        return this.frameOverlay != null;
+    }
+
+    public Matrix3f getRotationMatrix() {
+        return new Matrix3f();
+    }
+
+    public VoxelShape getOrCreateHitbox(Matrix3f rotationMatrix) {
```
Roelymole, line 234: It seems like all current uses of this method provide the value of the `getRotationMatrix` method above as the argument, so maybe you could streamline this a bit.

src/main/java/dev/galacticraft/mod/mixin/client/MultiPlayerGameModeMixin.java
```diff
@@ -0,0 +1,71 @@
...
+
+        VoxelShape wireShape = pipe.makeShape(wire);
+        VoxelShape frameShape = wire.getOrCreateHitbox(wire.getRotationMatrix());
+
+        boolean hitWire = preciseHit(wireShape, localHit);
+        boolean hitFrame = !hitWire && preciseHit(frameShape, localHit);
+
+        if (hitFrame) {
+            // Only break frame overlay
+            //level.setBlock(pos, wire.getOverlay().camouflage(), 3);
```
Roelymole, line 62: Do you need to keep this comment?

src/main/java/dev/galacticraft/mod/mixin/omnishape/ServerPlayerGameModeMixin.java
```diff
@@ -0,0 +1,78 @@
...
+            if (!player.gameMode.isCreative()) {
+                // Drop the frame
+                ItemStack frameStack = new ItemStack(OmnishapeBlocks.FRAME_BLOCK);
+                OmnishapeData.writeToItem(frameStack, wire.getOverlay());
+                Block.popResource(level, pos, frameStack);
+            }
+
+            // Remove the overlay but keep the wire
+            wire.setOverlay(null);
+            level.sendBlockUpdated(pos, wire.getBlockState(), wire.getBlockState(), 3);
```
Roelymole, line 73: Using `Block.UPDATE_ALL` instead of 3 is a bit more readable.

Roelymole (changes requested):

PR #499, opened 2025-08-06
Fixes some error,  specifically why blocks with a transparent RenderT…

Senthemodder:
so ...

The issue was that the oxygen bubble's visual effect, a semi-transparent blue sphere, was not being applied to blocks that were themselves transparent (like glass). This happened because of how Minecraft's rendering engine layers different types of graphics.

so  i added:
       - .setDepthTestState(new RenderStateShard.DepthTestStateShard("less", 513))
       -  .setTargetState(new RenderStateShard.TargetStateShard("translucent", () -> {}, () -> {}))

I also added a depth test to resolve some exiting errors in the shader. Should be good to go now.

### DIFF
--- a/src/main/java/dev/galacticraft/mod/Constant.java
+++ b/src/main/java/dev/galacticraft/mod/Constant.java
@@ -496,7 +496,7 @@ public interface Constant {
         String TITANIUM_HOE = "titanium_hoe";
 
         String STANDARD_WRENCH = "standard_wrench";
-        String GRAPPLING_HOOK = "grappling_hook";
+        // String GRAPPLING_HOOK = "grappling_hook";
         String TITANTIUM_UPGRADE_SMITHING_TEMPLATE = "titanium_upgrade_smithing_template";
         String BATTERY = "battery";
         String INFINITE_BATTERY = "infinite_battery";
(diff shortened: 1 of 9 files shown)


Roelymole:
Please could you move the grappling hook implementation to a separate PR since it is unrelated to the oxygen bubble bugfix. In general, it is much better to try to keep each PR focused on a single task. This will make it quicker and easier for your code to be reviewed and merged, as well as reducing the chance of something being overlooked. Also, I imagine the oxygen bubble bugfix will be ready much sooner than the grappling hook implementation, so it would be nice not to have to wait unnecessarily.

Also, if you haven't done already could familiarise yourself with the [contributing guidelines](https://github.com/TeamGalacticraft/Galacticraft/blob/main/.github/CONTRIBUTING.md). In particular, it would be great if you could try to use the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) style to write your commit messages and pull request titles, which should only be a small adjustment and take a few seconds extra.

Senthemodder:
I AM SO SORRY, I FORGOT TO COMMIT AND FELL ASLEEP OUT OF EXAUSTION, AAAAAA

Senthemodder:
Screenshot 2025-08-06 182720
I FELL ASLEEP AAAAAAAA....i hate myself...

Issue #493, opened 2025-07-02
Incorrect placement of tin ladder


MaverXka:
### Version Information

6d1c18a

### Reproduction steps
You can just place ladder anywhere
...

![Image](https://github.com/user-attachments/assets/25b8e378-10b9-4192-85c9-9f0be4e65f62)

ScottRox:
It is intentional. Tin ladders can be built on top of blocks and also underneath blocks. They also if looking down extend the ladders downwards if you keep clicking on it and upwards if looking upwards like scaffold builds upwards. when you keep clicking on it with more scaffold ion vanilla MC.

PR #502, opened 2025-08-11
Changes to Tin Ladder handling update

ScottRox:
Deals with defect Incorrect placement of tin ladder #493

Ladders now break when not supported. They can still be placed on and under solid blocks and extend up and down when looking up or down and placing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -219,3 +219,4 @@ gradle-app.setting
 run/
 remappedSrc/
 src/main/generated/.cache/
+/src/main/generated/
--- a/.gitignore
+++ b/.gitignore
@@ -219,4 +219,3 @@ gradle-app.setting
 run/
 remappedSrc/
 src/main/generated/.cache/
-/src/main/generated/
--- a/src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
+++ b/src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
@@ -40,6 +40,7 @@ import net.minecraft.world.level.block.state.BlockState;
 import net.minecraft.world.level.gameevent.GameEvent;
 import net.minecraft.world.phys.BlockHitResult;
 import org.jetbrains.annotations.Nullable;
+
 import java.util.Arrays;
 
 public class TinLadderBlock extends LadderBlock {
@@ -114,7 +115,7 @@ public class TinLadderBlock extends LadderBlock {
         Direction facingDirection = blockState.getValue(FACING);
 
         // Look at the block behind, above and below and check the appropriate face is sturdy or a ladder
-        for(Direction direction : Arrays.asList(facingDirection.getOpposite(), Direction.UP, Direction.DOWN) ) {
+        for (Direction direction : Arrays.asList(facingDirection.getOpposite(), Direction.UP, Direction.DOWN)) {
             BlockPos checkPos = blockPos.relative(direction);
             BlockState checkState = level.getBlockState(checkPos);
 
@@ -125,7 +126,7 @@ public class TinLadderBlock extends LadderBlock {
                 // If it's a tin ladder in facing the same way
                 if (checkState.is(GCBlocks.TIN_LADDER) && checkState.getValue(FACING) == facingDirection) return true;
             } else { // It's on the same level - not above or below.
-                // If the face is sturdy and it isn't a tin ladder. This prevents the situdation where a tin ladder can be placed
+                // If the face is sturdy and it isn't a tin ladder. This prevents the situation where a tin ladder can be placed
                 // back to back with another tin ladder and they can end up supporting themselves floating in the air.
                 if (checkState.isFaceSturdy(level, checkPos, direction.getOpposite()) && !checkState.is(GCBlocks.TIN_LADDER)) return true;
             }
@@ -133,8 +134,9 @@ public class TinLadderBlock extends LadderBlock {
         return false;
     }
 
-    /// This method looks to see if the ladder is supported by a solid block, of if it is attached to a ladder that is supported.
+    /// This method looks to see if the ladder is supported by a solid block, or if it's attached to a ladder that is supported.
     /// Using isFaceSturdy to determine if a block is solid as isSolid is deprecated and canOcclude doesn't give the desired result.
+    /// All ladders are not considered sturdy.
     /// It is recursive, so we have various termination scenarios before any recursion.
     public boolean isSupported(BlockState blockState, LevelReader level, BlockPos blockPos, Direction direction, Direction facingDirection) {
 
@@ -146,12 +148,12 @@ public class TinLadderBlock extends LadderBlock {
         // Change this to instanceof LadderBlock if you want wooden (and other ladders) to be considered support.
         if (!blockState.is(GCBlocks.TIN_LADDER)) return false;
 
-        // Check to ensure that ladder is facing in the same direction.
+        // Check to ensure that the ladder is facing in the same direction.
         if (blockState.getValue(FACING) != facingDirection) return false;
 
-        // Get the block behind the ladder - if the appropriate face is sturdy and its not a tin ladder then return true;
+        // Get the block behind the ladder - if the appropriate face is sturdy and it's not a ladder then return true;
         BlockState supportingState = level.getBlockState(blockPos.relative(blockState.getValue(FACING).getOpposite()));
-        if (supportingState.isFaceSturdy(level, blockPos, facingDirection) && !supportingState.is(GCBlocks.TIN_LADDER)) return true;
+        if (supportingState.isFaceSturdy(level, blockPos, facingDirection) && !(supportingState.getBlock() instanceof LadderBlock)) return true;
 
         // If we are moving in a particular vertical direction, then continue looking in that direction.
         if (direction.getAxis() == Direction.Axis.Y) {
--- a/src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
+++ b/src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
@@ -38,7 +38,6 @@ import net.minecraft.world.level.block.LadderBlock;
 import net.minecraft.world.level.block.state.BlockBehaviour;
 import net.minecraft.world.level.block.state.BlockState;
 import net.minecraft.world.level.gameevent.GameEvent;
-import net.minecraft.world.level.material.Fluids;
 import net.minecraft.world.phys.BlockHitResult;
 import org.jetbrains.annotations.Nullable;
 
--- a/src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
+++ b/src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
@@ -40,6 +40,7 @@ import net.minecraft.world.level.block.state.BlockState;
 import net.minecraft.world.level.gameevent.GameEvent;
 import net.minecraft.world.phys.BlockHitResult;
 import org.jetbrains.annotations.Nullable;
+import java.util.Arrays;
 
 public class TinLadderBlock extends LadderBlock {
     public TinLadderBlock(BlockBehaviour.Properties properties) {
@@ -49,6 +50,7 @@ public class TinLadderBlock extends LadderBlock {
     @Nullable
     private ItemInteractionResult checkCanTinLadderBePlaced(Level level, BlockPos checkPos, Player player, ItemStack itemStack, BlockState blockState) {
         BlockState checkState = level.getBlockState(checkPos);
+
         // Added check to ensure ladders are all facing the same way.
         if (checkState.is(GCBlocks.TIN_LADDER) && blockState.is(GCBlocks.TIN_LADDER) && checkState.getValue(FACING) != blockState.getValue(FACING)) {
             return ItemInteractionResult.PASS_TO_DEFAULT_BLOCK_INTERACTION;
@@ -93,8 +95,9 @@ public class TinLadderBlock extends LadderBlock {
         return super.useItemOn(stack, state, level, pos, player, hand, hit);
     }
 
-    // Need to override this method because Ladder Block is only interested in the block behind the ladder, which is useful functionality
-    // Be we want to pay attention to blocks above and below the ladders as well.
+    /// Need to override this method because Ladder Block is only interested in the block behind the ladder, which is useful
+    /// functionality, however, we want to pay attention to blocks above and below the ladders as well. This is handled
+    /// in the isSupported method.
     @Override
     protected BlockState updateShape(BlockState state, Direction direction, BlockState neighborState, LevelAccessor level, BlockPos pos, BlockPos neighborPos) {
         if (!isSupported(state, level, pos, direction.getOpposite(), state.getValue(FACING))) {
@@ -103,20 +106,41 @@ public class TinLadderBlock extends LadderBlock {
         return super.updateShape(state, direction, neighborState, level, pos, neighborPos);
     }
 
-    // Only called from the super updateShape class if the supporting block behind a ladder is broken.
+    /// Checks to see if a ladder could survive in the suggested location. Called when a nearby block is broken or
+    /// and sometimes when a new ladder is being placed.
     @Override
     public boolean canSurvive(BlockState blockState, LevelReader level, BlockPos blockPos) {
-        return isSupported(blockState, level, blockPos, blockState.getValue(FACING), blockState.getValue(FACING));
+        Direction facingDirection = blockState.getValue(FACING);
+
+        // Look at the block behind, above and below and check the appropriate face is sturdy or a ladder
+        for(Direction direction : Arrays.asList(facingDirection.getOpposite(), Direction.UP, Direction.DOWN) ) {
+            BlockPos checkPos = blockPos.relative(direction);
+            BlockState checkState = level.getBlockState(checkPos);
+
+            // If we are looking at the blocks above or below
+            if (direction == Direction.UP || direction == Direction.DOWN) {
+                // If the face is sturdy
+                if (checkState.isFaceSturdy(level, checkPos, direction.getOpposite())) return true;
+                // If it's a tin ladder in facing the same way
+                if (checkState.is(GCBlocks.TIN_LADDER) && checkState.getValue(FACING) == facingDirection) return true;
+            } else { // It's on the same level - not above or below.
+                // If the face is sturdy and it isn't a tin ladder. This prevents the situdation where a tin ladder can be placed
+                // back to back with another tin ladder and they can end up supporting themselves floating in the air.
+                if (checkState.isFaceSturdy(level, checkPos, direction.getOpposite()) && !checkState.is(GCBlocks.TIN_LADDER)) return true;
+            }
+        }
+        return false;
     }
 
-    // This method looks to see if the ladder is supported by a solid block, of if it is attached to a ladder that is supported.
-    // Using can occlude to determine if a block is solid as isSolid is deprecated.
-    // It is recursive, so we have various termination scenarios before any recursion.
+    /// This method looks to see if the ladder is supported by a solid block, of if it is attached to a ladder that is supported.
+    /// Using isFaceSturdy to determine if a block is solid as isSolid is deprecated and canOcclude doesn't give the desired result.
+    /// It is recursive, so we have various termination scenarios before any recursion.
     public boolean isSupported(BlockState blockState, LevelReader level, BlockPos blockPos, Direction direction, Direction facingDirection) {
         // If the current block we are looking at is solid, then the ladder is either supported underneath or above -> true
-        if (blockState.canOcclude()) return true;
+        if (blockState.isFaceSturdy(level, blockPos, Direction.UP)
+                || blockState.isFaceSturdy(level, blockPos, Direction.DOWN)) return true;
 
-        // If this isn't a tin ladder block and we know it isn't solid, then not supported.
+        // If this isn't a tin ladder block, and we know it isn't solid, then not supported.
         // Change this to instanceof LadderBlock if you want wooden (and other ladders) to be considered support.
         if (!blockState.is(GCBlocks.TIN_LADDER)) return false;
 
@@ -125,13 +149,14 @@ public class TinLadderBlock extends LadderBlock {
 
         // Get the block behind the ladder - if it can occlude (is solid) then return true;
         BlockState supportingState = level.getBlockState(blockPos.relative(blockState.getValue(FACING).getOpposite()));
-        if (supportingState.canOcclude()) return true;
+        if (supportingState.isFaceSturdy(level, blockPos, facingDirection)) return true;
 
         // If we are moving in a particular vertical direction, then continue looking in that direction.
-        if (direction.getAxis() == Direction.Axis.Y)
+        if (direction.getAxis() == Direction.Axis.Y) {
             return isSupported(level.getBlockState(blockPos.relative(direction)), level, blockPos.relative(direction), direction, facingDirection);
+        }
 
-        // If was horizontal direction, then need to look up and down to ensure find supported ladders there
+        // If was horizontal direction, then need to look up and down to find supported ladders there.
         return isSupported(level.getBlockState(blockPos.relative(Direction.UP)), level, blockPos.above(), Direction.UP, facingDirection)
                 || isSupported(level.getBlockState(blockPos.relative(Direction.DOWN)), level, blockPos.below(), Direction.DOWN, facingDirection);
     }
--- a/src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
+++ b/src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
@@ -117,9 +117,6 @@ public class TinLadderBlock extends LadderBlock {
         // If the current block we are looking at is solid, then the ladder is either supported underneath or above -> true
         if (blockState.canOcclude()) return true;
 
-        // If out of bounds, then not supported -> false
-        if (!insideBounds(blockPos, level)) return false;
-
         // If this isn't a tin ladder block and we know it isn't solid, then not supported.
         // Change this to instanceof LadderBlock if you want wooden (and other ladders) to be considered support.
         if (!blockState.is(GCBlocks.TIN_LADDER)) return false;
@@ -138,14 +135,5 @@ public class TinLadderBlock extends LadderBlock {
         // If was horizontal direction, then need to look up and down to ensure find supported ladders there
         return isSupported(level.getBlockState(blockPos.relative(Direction.UP)), level, blockPos.above(), Direction.UP, facingDirection)
                 || isSupported(level.getBlockState(blockPos.relative(Direction.DOWN)), level, blockPos.below(), Direction.DOWN, facingDirection);
-    }
-
-    // Returns true if the pos is inside the world/level boundaries.
-    public boolean insideBounds(BlockPos pos, LevelReader level) {
-        WorldBorder border = level.getWorldBorder();
-        return pos.getY() >= level.getMinBuildHeight() && pos.getY() <= level.getMaxBuildHeight()
-                && pos.getX() >= border.getMinX() && pos.getX() <= border.getMaxX()
-                && pos.getZ() >= border.getMinZ() && pos.getZ() <= border.getMaxZ();
-
     }
 }
--- a/src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
+++ b/src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
@@ -38,6 +38,7 @@ import net.minecraft.world.level.block.LadderBlock;
 import net.minecraft.world.level.block.state.BlockBehaviour;
 import net.minecraft.world.level.block.state.BlockState;
 import net.minecraft.world.level.gameevent.GameEvent;
+import net.minecraft.world.level.material.Fluids;
 import net.minecraft.world.phys.BlockHitResult;
 import org.jetbrains.annotations.Nullable;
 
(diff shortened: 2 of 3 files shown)


src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
```diff
@@ -82,8 +91,59 @@ protected ItemInteractionResult useItemOn(ItemStack stack, BlockState state, Lev
         return super.useItemOn(stack, state, level, pos, player, hand, hit);
     }
 
+    // need to override this method because LAddBlock is only interested in the block behind the ladder, which is useful functionality
```
Roelymole: LAddBlock?

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
```diff
@@ -82,8 +91,59 @@ protected ItemInteractionResult useItemOn(ItemStack stack, BlockState state, Lev
...
-        return true;
+    @NotNull
+    protected BlockState updateShape(BlockState state, Direction axisDirection, BlockState neighborState, LevelAccessor level, BlockPos pos, BlockPos neighborPos)
+    {
+        if(!isSupported(state, level, pos, axisDirection.getOpposite()))
+            return Blocks.AIR.defaultBlockState();
+        return super.updateShape(state, axisDirection, neighborState, level, pos, neighborPos);
+    }
+
+    // this method never gets called due to the way
```
Roelymole: Due to the way what...?

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
```diff
@@ -82,8 +91,59 @@ protected ItemInteractionResult useItemOn(ItemStack stack, BlockState state, Lev
...
 
+    // need to override this method because LAddBlock is only interested in the block behind the ladder, which is useful functionality
+    // be we want to pay attention to blocks above and below the ladders as well.
     @Override
-    public boolean canSurvive(BlockState blockState, LevelReader level, BlockPos blockPos) {
-        return true;
+    @NotNull
+    protected BlockState updateShape(BlockState state, Direction axisDirection, BlockState neighborState, LevelAccessor level, BlockPos pos, BlockPos neighborPos)
+    {
+        if(!isSupported(state, level, pos, axisDirection.getOpposite()))
```
Roelymole: We have been putting a space between the if keyword and the opening bracket throughout the rest of the codebase, so please could you do the same here (and in each other instance below where you have done this) in order to maintain consistency, thanks!

I would also prefer enclosing the body of this if statement in curly brackets since the return statement is on a separate line.

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
```diff
@@ -82,8 +91,59 @@ protected ItemInteractionResult useItemOn(ItemStack stack, BlockState state, Lev
...
     }
 
+    // need to override this method because LAddBlock is only interested in the block behind the ladder, which is useful functionality
+    // be we want to pay attention to blocks above and below the ladders as well.
     @Override
-    public boolean canSurvive(BlockState blockState, LevelReader level, BlockPos blockPos) {
-        return true;
+    @NotNull
+    protected BlockState updateShape(BlockState state, Direction axisDirection, BlockState neighborState, LevelAccessor level, BlockPos pos, BlockPos neighborPos)
+    {
```
Roelymole: The bracketing style we are using is to put the curly bracket on the same line rather than on a new line, as mentioned in the [code style guidelines](https://github.com/TeamGalacticraft/Galacticraft/blob/main/.github/CONTRIBUTING.md#code-style).

src/main/resources/fabric.mod.json
```diff
@@ -23,6 +23,12 @@
     }
   ],
   "contributors": [
+    {
+      "name": "ScottRox",
+      "contact": {
+        "homepage": "https://github.com/ScottRox"
+      }
+    },
```
Roelymole: I think the order of contributors listed here is based on the order in which they appear on GitHub. Although this is fairly outdated, so perhaps leave it out of this pull request and we can see about updating them all properly - ideally in an automated, future-proof way somehow.

By no means am I suggesting denying you recognition.

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
```diff
@@ -82,8 +91,59 @@ protected ItemInteractionResult useItemOn(ItemStack stack, BlockState state, Lev
...
+    }
+
+    // returns true if the pos is inside the world/level boundaries.
+    public boolean insideBounds(BlockPos pos, LevelReader level)
+    {
+        WorldBorder border = level.getWorldBorder();
+        return pos.getY() >= level.getMinBuildHeight() && pos.getY() <= level.getMaxBuildHeight()
+                && pos.getX() >= border.getMinX() && pos.getX() <= border.getMaxX()
+                && pos.getZ() >= border.getMinZ() && pos.getZ() <= border.getMaxZ();
+
```
Roelymole, line 165: Could you remove this blank line.

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
```diff
@@ -82,8 +91,59 @@ protected ItemInteractionResult useItemOn(ItemStack stack, BlockState state, Lev
         return super.useItemOn(stack, state, level, pos, player, hand, hit);
     }
 
+    // need to override this method because LAddBlock is only interested in the block behind the ladder, which is useful functionality
+    // be we want to pay attention to blocks above and below the ladders as well.
```
Roelymole: I don't think the word 'be' makes much sense at the beginning of this comment. Do you perhaps mean because?

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
```diff
@@ -82,8 +91,59 @@ protected ItemInteractionResult useItemOn(ItemStack stack, BlockState state, Lev
...
+        // if out of bounds, then not supported -> false
+        if(!insideBounds(blockPos, level)) return false;
+
+        // if this isn't a tin ladder block and we know it isn't solid, then not supported.
+        // change this to instanceof LadderBlock if you want wooden (and other ladders) to be considered support.
+        if(!blockState.is(GCBlocks.TIN_LADDER)) return false;
+
+        // get the block behind the ladder - if it can occlude (is solid) then return true;
+        BlockState supportingState = level.getBlockState(blockPos.relative(blockState.getValue(FACING).getOpposite()));
+        if(supportingState.canOcclude()) return true;
```
Roelymole: It would be better to use the `canAttachTo` method from the parent class `LadderBlock`, which is what the `canSurvive` method for vanilla ladders uses, rather than `canOcclude`. For example, something like the following:

```
if (this.canAttachTo(level, blockPos.relative(direction.getOpposite()), direction)) return true;
```

Otherwise tin ladders can't be placed on the back of open trapdoors, whereas vanilla ladders can.

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
```diff
@@ -82,8 +91,59 @@ protected ItemInteractionResult useItemOn(ItemStack stack, BlockState state, Lev
...
+        return isSupported(blockState, level, blockPos, blockState.getValue(FACING));
+    }
+
+
+    // this method looks to see if the ladder is supported by a solid block, of if it is attached to a ladder that is supported.
+    // Using can occlude to determine if a block is solid as isSolid is deprecated.
+    public boolean isSupported(BlockState blockState, LevelReader level, BlockPos blockPos, Direction direction)
+    {
+        // if the current block we are looking at is solid then ladder is either supported underneath or above -> true
+        if(blockState.canOcclude()) return true;
```
Roelymole: I'm not sure if using `canOcclude` is quite appropriate, because I believe this line is what allows for situations like this.

Image

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
```diff
@@ -82,8 +91,59 @@ protected ItemInteractionResult useItemOn(ItemStack stack, BlockState state, Lev
         return super.useItemOn(stack, state, level, pos, player, hand, hit);
     }
 
+    // need to override this method because LAddBlock is only interested in the block behind the ladder, which is useful functionality
+    // be we want to pay attention to blocks above and below the ladders as well.
     @Override
-    public boolean canSurvive(BlockState blockState, LevelReader level, BlockPos blockPos) {
-        return true;
+    @NotNull
+    protected BlockState updateShape(BlockState state, Direction axisDirection, BlockState neighborState, LevelAccessor level, BlockPos pos, BlockPos neighborPos)
```
Roelymole: `axisDirection` should probably just be named `direction`.

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
```diff
@@ -82,8 +91,59 @@ protected ItemInteractionResult useItemOn(ItemStack stack, BlockState state, Lev
...
+
+    // this method looks to see if the ladder is supported by a solid block, of if it is attached to a ladder that is supported.
+    // Using can occlude to determine if a block is solid as isSolid is deprecated.
+    public boolean isSupported(BlockState blockState, LevelReader level, BlockPos blockPos, Direction direction)
+    {
+        // if the current block we are looking at is solid then ladder is either supported underneath or above -> true
+        if(blockState.canOcclude()) return true;
+
+        // if out of bounds, then not supported -> false
+        if(!insideBounds(blockPos, level)) return false;
```
Roelymole: Is this necessary at all? From what I can see this doesn't do anything different from normal, but I might be missing something.

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
```diff
@@ -82,8 +91,59 @@ protected ItemInteractionResult useItemOn(ItemStack stack, BlockState state, Lev
...
+
+        // if this isn't a tin ladder block and we know it isn't solid, then not supported.
+        // change this to instanceof LadderBlock if you want wooden (and other ladders) to be considered support.
+        if(!blockState.is(GCBlocks.TIN_LADDER)) return false;
+
+        // get the block behind the ladder - if it can occlude (is solid) then return true;
+        BlockState supportingState = level.getBlockState(blockPos.relative(blockState.getValue(FACING).getOpposite()));
+        if(supportingState.canOcclude()) return true;
+
+        // if we are moving in a particlar vertical direction then continue looking in that direction.
```
Roelymole: Particular is misspelt.

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
```diff
@@ -82,8 +91,59 @@ protected ItemInteractionResult useItemOn(ItemStack stack, BlockState state, Lev
...
+    // Only called from the super updateShape class if the supporting block behind a ladder is broken.
+    @Override
+    public boolean canSurvive(BlockState blockState, LevelReader level, BlockPos blockPos)
+    {
+        return isSupported(blockState, level, blockPos, blockState.getValue(FACING));
+    }
+
+
+    // this method looks to see if the ladder is supported by a solid block, of if it is attached to a ladder that is supported.
+    // Using can occlude to determine if a block is solid as isSolid is deprecated.
```
Roelymole: Could you be consistent in the capitalisation at the start of comments?

I would tend towards capitalising the first word myself.

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
```diff
@@ -61,16 +67,19 @@ private ItemInteractionResult checkCanTinLadderBePlaced(Level level, BlockPos ch
     }
 
     @Override
+    @NotNull
```
Roelymole: Other classes that override the `useItemOn` method do not use the `@NotNull` decorator, so it probably isn't necessary here either.

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
```diff
@@ -82,8 +91,59 @@ protected ItemInteractionResult useItemOn(ItemStack stack, BlockState state, Lev
...
+    }
+
+    // Only called from the super updateShape class if the supporting block behind a ladder is broken.
+    @Override
+    public boolean canSurvive(BlockState blockState, LevelReader level, BlockPos blockPos)
+    {
+        return isSupported(blockState, level, blockPos, blockState.getValue(FACING));
+    }
+
+
```
Roelymole, line 149: There is an extra blank line here too.

src/main/resources/fabric.mod.json
ScottRox: NP

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
ScottRox: Just used the same signature as the method it is overriding. Changed it though.

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
ScottRox: Assuming you're talking about the insideBounds bit. I'm not sure. I put it in to be safe. Will test to see what happens if not there.

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
ScottRox: NP

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
ScottRox: no idea how that happened.

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
ScottRox: NP

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
ScottRox: NP

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
ScottRox: Neither does updateShape. Removing from both. Was just following idea suggestions...

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
ScottRox: NP

ScottRox:
Yeah. It's a recursive routine. If a ladders supporting block is broken, then it looks up and down to see if there are supporting blocks or supported ladders.

ScottRox:
...supported tin ladders. I don't actually check they are facing the same direction though. Perhaps that would be worth doing too.

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
Roelymole: Yeah, that is what I was meaning.

Roelymole:
I mean does it check that the `FACING` block state property matches?

ScottRox:
Going to fix that. It just looks weird.

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
ScottRox: Removed.

.gitignore
```diff
@@ -219,3 +219,4 @@ gradle-app.setting
 run/
 remappedSrc/
 src/main/generated/.cache/
+/src/main/generated/
```
marcus8448: Accidental change? (this isn't an unused import)
We want the generated assets tracked in git.

ScottRox:
So I think I've done everything, including the facing the right was stuff. Do I keep working on new stuff in my fork, or should I make another fork? It doesn't look like the pull request has been approved.

Roelymole:
You don't need to create a new fork, but it is best to create a new branch. There are several ways you can go about this, none of which are very difficult.

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
Roelymole: Using `blockState.isFaceSturdy(level, blockPos, Direction.UP)` might be a good alternative.

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
ScottRox: Fixed

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
ScottRox: Cool. Will test that option.

ScottRox: Resolved. Sorry, I didn't see this before.

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
ScottRox: This leads us to a problem where the back of a tin ladder can support another tin ladder. You can then break the original ladders supporting blocks and the two ladders will float back to back. The only way I can think to resolve this situation is to explicitly code for not allowing tin ladders to be placed on the back of other tin ladders, which I've done.

.gitignore
ScottRox: Okay. Thought I didn't commit .gitignore, but will resolve.

ScottRox: Removex that line from .gitignore

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
```diff
@@ -82,8 +95,71 @@ protected ItemInteractionResult useItemOn(ItemStack stack, BlockState state, Lev
         return super.useItemOn(stack, state, level, pos, player, hand, hit);
     }
 
+    /// Need to override this method because Ladder Block is only interested in the block behind the ladder, which is useful
```
Roelymole, line 103: Is there a reason why you are using 3 forward slashes for some comments?

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
```diff
@@ -82,8 +95,71 @@ protected ItemInteractionResult useItemOn(ItemStack stack, BlockState state, Lev
...
+
+            // If we are looking at the blocks above or below
+            if (direction == Direction.UP || direction == Direction.DOWN) {
+                // If the face is sturdy
+                if (checkState.isFaceSturdy(level, checkPos, direction.getOpposite())) return true;
+                // If it's a tin ladder in facing the same way
+                if (checkState.is(GCBlocks.TIN_LADDER) && checkState.getValue(FACING) == facingDirection) return true;
+            } else { // It's on the same level - not above or below.
+                // If the face is sturdy and it isn't a tin ladder. This prevents the situdation where a tin ladder can be placed
+                // back to back with another tin ladder and they can end up supporting themselves floating in the air.
```
Roelymole: Situation is misspelt, but also it seems that the same thing can still occur if one of the ladders is a normal ladder.

Image

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
Roelymole: This extra blank line is still here.

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
```diff
@@ -82,8 +95,71 @@ protected ItemInteractionResult useItemOn(ItemStack stack, BlockState state, Lev
...
+                || (blockState.isFaceSturdy(level, blockPos, Direction.DOWN) && (direction == Direction.UP))) return true;
+
+        // If this isn't a tin ladder block, and we know it isn't solid, then not supported.
+        // Change this to instanceof LadderBlock if you want wooden (and other ladders) to be considered support.
+        if (!blockState.is(GCBlocks.TIN_LADDER)) return false;
+
+        // Check to ensure that ladder is facing in the same direction.
+        if (blockState.getValue(FACING) != facingDirection) return false;
+
+        // Get the block behind the ladder - if the appropriate face is sturdy and its not a tin ladder then return true;
```
Roelymole: It should really be "it's" not "its" since it is a contraction of "it is".

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
```diff
@@ -22,20 +22,25 @@
...
+import net.minecraft.world.level.LevelAccessor;
 import net.minecraft.world.level.LevelReader;
+import net.minecraft.world.level.block.Blocks;
 import net.minecraft.world.level.block.LadderBlock;
 import net.minecraft.world.level.block.state.BlockBehaviour;
 import net.minecraft.world.level.block.state.BlockState;
 import net.minecraft.world.level.gameevent.GameEvent;
 import net.minecraft.world.phys.BlockHitResult;
 import org.jetbrains.annotations.Nullable;
+import java.util.Arrays;
```
Roelymole, line 44: It is the convention throughout the whole project to put the `java.util` imports at the bottom separated from the rest of the imports with a blank line.

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
```diff
@@ -103,20 +106,41 @@ protected BlockState updateShape(BlockState state, Direction direction, BlockSta
...
-    // Only called from the super updateShape class if the supporting block behind a ladder is broken.
+    /// Checks to see if a ladder could survive in the suggested location. Called when a nearby block is broken or
+    /// and sometimes when a new ladder is being placed.
     @Override
     public boolean canSurvive(BlockState blockState, LevelReader level, BlockPos blockPos) {
-        return isSupported(blockState, level, blockPos, blockState.getValue(FACING), blockState.getValue(FACING));
+        Direction facingDirection = blockState.getValue(FACING);
+
+        // Look at the block behind, above and below and check the appropriate face is sturdy or a ladder
+        for(Direction direction : Arrays.asList(facingDirection.getOpposite(), Direction.UP, Direction.DOWN) ) {
```
Roelymole: The use of spaces isn't quite right here.

```suggestion
        for (Direction direction : Arrays.asList(facingDirection.getOpposite(), Direction.UP, Direction.DOWN)) {
```

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
```diff
@@ -44,7 +49,13 @@ public TinLadderBlock(BlockBehaviour.Properties properties) {
...
     private ItemInteractionResult checkCanTinLadderBePlaced(Level level, BlockPos checkPos, Player player, ItemStack itemStack, BlockState blockState) {
-        if (level.getBlockState(checkPos).canBeReplaced()) {
+        BlockState checkState = level.getBlockState(checkPos);
+
+        // Added check to ensure ladders are all facing the same way.
+        if (checkState.is(GCBlocks.TIN_LADDER) && blockState.is(GCBlocks.TIN_LADDER) && checkState.getValue(FACING) != blockState.getValue(FACING)) {
+            return ItemInteractionResult.PASS_TO_DEFAULT_BLOCK_INTERACTION;
+        }
+        if (checkState.canBeReplaced()) {
             var newState = this.defaultBlockState().setValue(FACING, blockState.getValue(FACING));
```
Roelymole: The `WATERLOGGED` block state property should be set to true if where the ladder is to be placed is currently a water source.

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
ScottRox: Nothing significant. In intellij it makes the comments look nicer before the functions. inline doesn't seem to make a difference. It does color them green though. You can also use html or maybe just simple mark up for the comments. I did it on a whim. Doesn't need to be there.

ScottRox: Do you want me to remove it or leave it?

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
ScottRox: NP

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
Roelymole: Ah, okay. It isn't a big deal either way. I was curious as to why more than anything.

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
ScottRox: Handled by the parent class. Behaviour is consistent with wooden ladders.

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
ScottRox: I'm not seeing it. It isn't in github either from what I can see.

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
ScottRox: image

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
ScottRox: Check to ensure ladders cannot support ladders instead of specifically for tin ladders.

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
```diff
@@ -82,8 +96,72 @@ protected ItemInteractionResult useItemOn(ItemStack stack, BlockState state, Lev
...
+        }
+        return false;
+    }
+
+    /// This method looks to see if the ladder is supported by a solid block, or if it's attached to a ladder that is supported.
+    /// Using isFaceSturdy to determine if a block is solid as isSolid is deprecated and canOcclude doesn't give the desired result.
+    /// All ladders are not considered sturdy.
+    /// It is recursive, so we have various termination scenarios before any recursion.
+    public boolean isSupported(BlockState blockState, LevelReader level, BlockPos blockPos, Direction direction, Direction facingDirection) {
+
```
Roelymole: Maybe this is the blank line I meant. I guess it isn't too surprising that something so generic gets muddled up.

Roelymole: Just below `public boolean isSupported(...) {` to be clear.

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
Roelymole: Maybe I wasn't quite clear, so here is a screen recording that should show you exactly what I mean.

https://github.com/user-attachments/assets/d917121d-bc16-4f34-855c-cc56b7c0b47d

Placing a tin ladder in water directly is fine, it is when the ladder is extended into a water source that isn't quite right.

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
ScottRox: Resolving this.

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
ScottRox: Fixed.

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
ScottRox: Fixed.

src/main/java/dev/galacticraft/mod/content/block/special/TinLadderBlock.java
```diff
@@ -57,7 +57,9 @@ private ItemInteractionResult checkCanTinLadderBePlaced(Level level, BlockPos ch
             return ItemInteractionResult.PASS_TO_DEFAULT_BLOCK_INTERACTION;
         }
         if (checkState.canBeReplaced()) {
-            var newState = this.defaultBlockState().setValue(FACING, blockState.getValue(FACING));
+            var newState = this.defaultBlockState()
+                    .setValue(FACING, blockState.getValue(FACING))
+                    .setValue(WATERLOGGED, checkState.is(Blocks.WATER) || checkState.getValue(WATERLOGGED));
```
Roelymole: You should probably do something more like this so that only water sources and full blocks of water are converted.

```suggestion
            boolean wasWater = checkState.is(Blocks.WATER) && checkState.getFluidState().getAmount() >= 8 && check.getFluidState().isSource();
            var newState = this.defaultBlockState()
                    .setValue(FACING, blockState.getValue(FACING))
                    .setValue(WATERLOGGED, wasWater || checkState.getValue(WATERLOGGED));
```

We also need to be more careful when using `checkState.getValue(WATERLOGGED)`.

```
java.lang.IllegalArgumentException: Cannot get property BooleanProperty{name=waterlogged, clazz=class java.lang.Boolean, values=[true, false]} as it does not exist in Block{minecraft:air}
```

You could use either `checkState.hasProperty(WATERLOGGED)` or `checkState.getOptionalValue(WATERLOGGED)`.

ScottRox: Modified this a bit. Check my logic. Not sure what circumstances you were testing, but passes the ones I checked.

Roelymole: That seems good now, thanks!

Roelymole (approved):
I think this is all good now, thanks!

ScottRox:
So in regard to the web string and torch... I'm ready to submit that too. Do I just merge my branch back into the fork and do another pull request?

PR #503, opened 2025-08-12
Contributors in fabric.mod.json overhaul

maxryan008:
Contributors are now auto downloaded from Github when fabric loom tasks are run. If contributions are unchanged then download is skipped.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -22,6 +22,14 @@
 
 import java.time.LocalDateTime
 import java.time.format.DateTimeFormatter
+import com.google.gson.Gson
+import com.google.gson.GsonBuilder
+import com.google.gson.JsonElement
+import com.google.gson.JsonObject
+import com.google.gson.JsonParser
+import com.google.gson.reflect.TypeToken
+import java.net.HttpURLConnection
+import java.net.URL
 
 // Build Info
 val isCi = (System.getenv("CI") ?: "false") == "true"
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -218,6 +226,183 @@ configurations {
...
+val etagFile = file("$buildDir/contributors.etag")
+
+val gsonPretty: Gson = GsonBuilder().setPrettyPrinting().disableHtmlEscaping().create()
+val gson: Gson = GsonBuilder().disableHtmlEscaping().create()
+
+data class ContributorApi(
+    val login: String?,
+    val html_url: String?,
+    val type: String?,
+    val contributions: Int?
```
marcus8448: The number of contributions shouldn't be put in the FMJ. It's not in the spec and will cause a lot of churn.

marcus8448 (changes requested):
I think this should be packaged in a GitHub action (or at least a Gradle plugin), and run on a cron job (or per-release, when that happens) instead of on every build/run. Otherwise, if you checkout an old commit or branch (or someone contributes while you're working), building/running will introduce unrelated `fabric.mod.json` changes.

build.gradle.kts
maxryan008: ok i can remove number of contributions and i will look into the cron job / github actions system

.github/scripts/update_contributors.py
```diff
@@ -0,0 +1,123 @@
...
+            first = False
+
+    return results, latest_etag
+
+def build_payload(raw):
+    # filter bots & sort by contributions desc
+    filtered = [
+        c for c in raw
+        if not ((c.get("type") or "").lower().__contains__("bot")
+                or (c.get("login") or "").lower().__contains__("bot"))
```
Roelymole: It isn't great practice to call the `__contains__` dunder method directly.

The syntax `a in b` invokes `b.__contains__(a)` internally.

Roelymole: Using the syntax `dict.get(key, fallback)` might be neater than `dict.get(key) or fallback` in this case.

For example, `"bot" in c.get("type", "").lower()` compared to `"bot" in (c.get("type") or "").lower()`.

.github/scripts/update_contributors.py
```diff
@@ -0,0 +1,123 @@
...
+
+    new_contrib = build_payload(raw)
+
+    if existing == new_contrib:
+        print("Contributors already up-to-date. No changes written.")
+        if new_etag:
+            os.makedirs(os.path.dirname(ETAG_PATH), exist_ok=True)
+            with open(ETAG_PATH, "w", encoding="utf-8") as f:
+                f.write(new_etag)
+        return
```
Roelymole: I think this could be refactored to avoid repeating the `if new_etag:` block of code below, in the following way:

```
    if new_etag:
        os.makedirs(os.path.dirname(ETAG_PATH), exist_ok=True)
        with open(ETAG_PATH, "w", encoding="utf-8") as f:
            f.write(new_etag)

    if existing == new_contrib:
        print("Contributors already up-to-date. No changes written.")
        return
```

.github/workflows/update_contributors.yml
```diff
@@ -0,0 +1,51 @@
+name: ContributorsCheck
+
+on:
+  push:
+    branches:
+      - '**'
+      - '!chore/update-contributors'   # avoid loops from bot updates
+  workflow_dispatch: {}
```
marcus8448: I still think this should be a cron - or at least run only on main. Leaving workflow_dispatch is fine though.

marcus8448: Since the action directly pushes a commit, if anyone pushes code on a branch that isn't up to date it would cause duplicate commits to be generated.

marcus8448 (approved):

PR #504, opened 2025-08-12
Config screen overhaul

maxryan008:
Made all config options available Made new categories and made small category not have subcategories. Config text no longer overflows into the text box area.

### DIFF
--- a/src/main/java/dev/galacticraft/mod/config/ConfigImpl.java
+++ b/src/main/java/dev/galacticraft/mod/config/ConfigImpl.java
@@ -29,6 +29,7 @@ import dev.galacticraft.mod.Galacticraft;
 import dev.galacticraft.mod.api.config.Config;
 import dev.galacticraft.mod.util.Translations;
 import me.shedaniel.clothconfig2.api.ConfigBuilder;
+import me.shedaniel.clothconfig2.api.ConfigCategory;
 import me.shedaniel.clothconfig2.api.ConfigEntryBuilder;
 import me.shedaniel.clothconfig2.impl.builders.BooleanToggleBuilder;
 import me.shedaniel.clothconfig2.impl.builders.DoubleFieldBuilder;
@@ -36,8 +37,11 @@ import me.shedaniel.clothconfig2.impl.builders.FloatFieldBuilder;
 import me.shedaniel.clothconfig2.impl.builders.LongFieldBuilder;
 import me.shedaniel.clothconfig2.impl.builders.SubCategoryBuilder;
 import net.fabricmc.fabric.api.transfer.v1.fluid.FluidConstants;
+import net.minecraft.client.gui.Font;
 import net.minecraft.client.gui.screens.Screen;
 import net.minecraft.network.chat.Component;
+import net.minecraft.network.chat.MutableComponent;
+import org.jetbrains.annotations.Nullable;
 
 import java.io.File;
 import java.io.FileReader;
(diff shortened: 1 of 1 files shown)


src/main/java/dev/galacticraft/mod/config/ConfigImpl.java
```diff
@@ -429,52 +433,81 @@ public Screen create(Screen parent) {
                     .setTitle(Component.translatable(Translations.Config.TITLE))
                     .setSavingRunnable(config::save);
 
-            SubCategoryBuilder dB = ConfigEntryBuilder.create().startSubCategory(Component.translatable(Translations.Config.DEBUG));
+            var font = net.minecraft.client.Minecraft.getInstance().font;
+            final int MAX_LW = computeMaxLabelWidthPx();
 
-            dB.add(new BooleanToggleBuilder(
+            java.util.function.Function<String, Component> L =
```
Roelymole: Could you add this to the imports at the top, please.

src/main/java/dev/galacticraft/mod/config/ConfigImpl.java
```diff
@@ -641,141 +691,189 @@ public Screen create(Screen parent) {
...
             );
 
-            b.getOrCreateCategory(Component.translatable(Translations.Config.DEBUG)).addEntry(dB.build());
-            b.getOrCreateCategory(Component.translatable(Translations.Config.ENERGY)).addEntry(wires.build()).addEntry(machines.build());
+            return b.build();
+        }
 
-            SubCategoryBuilder commands = ConfigEntryBuilder.create().startSubCategory(Component.translatable(Translations.Config.COMMANDS));
+        private Component buildTooltip(MutableComponent name, Font font, int maxLw, @Nullable MutableComponent desc) {
+            var fcs = name.getVisualOrderText();
```
Roelymole: It isn't clear (to me at least) what `fcs` stands for, especially when `var` is used so there isn't a type to hint at what this is either.

src/main/java/dev/galacticraft/mod/config/ConfigImpl.java
```diff
@@ -641,141 +691,189 @@ public Screen create(Screen parent) {
...
+            String s = full.getString();
+            int ell = font.width("…");
+            if (ell >= maxPx) return Component.literal("");
+
+            String cut = font.plainSubstrByWidth(s, maxPx - ell);
+            return Component.literal(cut + "…");
+        }
+
+        private static int computeMaxLabelWidthPx() {
+            var win = net.minecraft.client.Minecraft.getInstance().getWindow();
```
Roelymole: This is the second time you are using the full path, so you should definitely just add an import. It looks much cleaner too.

src/main/java/dev/galacticraft/mod/config/ConfigImpl.java
```diff
@@ -429,52 +433,81 @@ public Screen create(Screen parent) {
...
+                            Component.translatable(id),
+                            font,
+                            MAX_LW,
+                            (optDescId != null && !optDescId.isEmpty())
+                                    ? Component.translatable(optDescId)
+                                    : null
+                    );
+
+            java.util.function.Function<String, Component> TT1 =
+                    id -> TT.apply(id, null);
```
Roelymole: These function names `L`, `TT` and `TT1` aren't very descriptive. Could you think of something else, or at least add a brief comment explaining what they stand for.

src/main/java/dev/galacticraft/mod/config/ConfigImpl.java
```diff
@@ -429,52 +433,81 @@ public Screen create(Screen parent) {
                     .setTitle(Component.translatable(Translations.Config.TITLE))
                     .setSavingRunnable(config::save);
 
-            SubCategoryBuilder dB = ConfigEntryBuilder.create().startSubCategory(Component.translatable(Translations.Config.DEBUG));
+            var font = net.minecraft.client.Minecraft.getInstance().font;
```
Roelymole: There is not much reason to use `var font` rather than `Font font` when you are already importing Font and it is only one character longer.

src/main/java/dev/galacticraft/mod/config/ConfigImpl.java
```diff
@@ -641,141 +691,189 @@ public Screen create(Screen parent) {
...
+            int ell = font.width("…");
+            if (ell >= maxPx) return Component.literal("");
+
+            String cut = font.plainSubstrByWidth(s, maxPx - ell);
+            return Component.literal(cut + "…");
+        }
+
+        private static int computeMaxLabelWidthPx() {
+            var win = net.minecraft.client.Minecraft.getInstance().getWindow();
+            int screenW = win.getGuiScaledWidth();
```
Roelymole: Since `win` is only being used once, there isn't necessarily much point storing it in a variable on the line before.

Roelymole (changes requested):
There are a couple little stylistic changes I would prefer to see.

src/main/java/dev/galacticraft/mod/config/ConfigImpl.java
maxryan008: i can drop some comments. its specifically L stands for Label, TT stands for tool tip and TT1 it Tooltip but 1 argument

src/main/java/dev/galacticraft/mod/config/ConfigImpl.java
```diff
@@ -429,52 +437,103 @@ public Screen create(Screen parent) {
...
+            // Use for normal Labels under a category.
+            Function<String, Component> label =
+                    key -> ellipsize(Component.translatable(key), font, maxLabelWidth);
+
+            // Use for labels under a subcategory.
+            Function<String, Component> labelSub =
+                    key -> ellipsize(Component.translatable(key), font, maxLabelWidth - 15);
+
+            // Use for tooltips with description e.g. (name, description)
+            BiFunction<String, String, Component> toolTipWithDesc =
```
Roelymole: I think the capitalisation `tooltipWithDesc` would be better and similarly `tooltipSingular`, `tooltipWithDescSub` and `tooltipSingularSub` for the other functions. Especially since there is the `buildTooltip` method already using this style of capitalisation.

Roelymole (approved):
Wonderful job!

PR #508, opened 2025-08-15
feat: allow REI and JEI to be selected easily

Roelymole:
Allow REI and JEI to be selected easily when running Galacticraft in a development environment by providing additional arguments to gradlew.

`-Prei=true/false (default: true)`
`-Pjei=true/false (default: false)`

For example, `.\gradlew runClient -Prei=false` would disable both recipe viewers and `.\gradlew.runClient -Pjei=true` would enable both recipe viewers.

To use a single recipe viewer, use the command `.\gradlew runClient -Prei=false -Pjei=true` or `.\gradlew runClient -Prei=true -Pjei=false` (equivalent to `.\gradlew runClient`).


build.gradle.kts
```diff
@@ -25,6 +25,8 @@ import java.time.format.DateTimeFormatter
 
 // Build Info
 val isCi = (System.getenv("CI") ?: "false") == "true"
+val runRei = project.getProperties().getOrDefault("rei", "true").toString().toBoolean()
+val runJei = project.getProperties().getOrDefault("jei", "false").toString().toBoolean()
```
marcus8448: Is there a usecase for loading both? Otherwise it might be nicer to just have a JEI toggle (that disables REI) and just check for the presence of a JEI property (`project.properties.containsKey("jei")`).

marcus8448 (approved):

build.gradle.kts
Roelymole: Umm, it might be helpful when checking for parity, but having one or the other is definitely the more common usecase. I'll have a think about it and try to come up with something more streamlined (that should be able to accomodate EMI as well, in the future).

Roelymole: How is that? Hopefully it isn't too confusing...

marcus8448: Looks good to me - it's just a dev tool so it doesn't matter that much if it's not the prettiest

PR #509, opened 2025-08-16
feat: add recipe viewers for canning foods

Roelymole:
Full support for viewing how to make canned foods in the food canner using REI/JEI and partial support for EMI (only the default items in the canned food creative tab have recipes, since EMI does not allow for dynamic recipes).

- [x] Animate the progress bars and show the whole process with a single can moving from the top slot to the bottom slot.
- [x] The canning tab doesn't currently appear in the list of all categories since there aren't any recipe data files for canning and instead the list of recipes is dynamically generated.
- [x] Add support for the move items button. *(Doesn't currently transfer an empty can to the top slot, which might be something worth adding.)*
- [ ] Use custom transfer handlers to allow the move items button in REI/JEI to transfer more than one item into a slot.
- [ ] Handle displaying more than 16 distinct items (not currently possible as the `MAX_FOOD` value is hardcoded, but necessary if we make this value configurable).

An example of the recipe viewer for a can with 16 different foods (by default this can be seen by pressing R when hovering over the item).

---

The recipes for all the canned foods consisting of a full can of a single item are listed when viewing the usage of the food canner (by default this can be seen by pressing U when hovering over the item). These are the canned foods listed in REI/JEI on the right and also the canned foods creative mode tab.

### JEI

---

### REI

https://github.com/user-attachments/assets/df3d76c7-47b8-4d61-8e32-926275dba723

### JEI

https://github.com/user-attachments/assets/2fdc92e2-dcdc-4d86-b747-ba3e8d80faf7


marcus8448 (changes requested):
A couple bugs:


The transfer handler items seem to be misaligned with the rendered ones (REI):

image

---

There's an empty can canning recipe (both REI and JEI), if you place an empty can into a canner and view its recipe when it's in the 'processing' slot.
image

---

If you look at the 'uses' of an empty can, the items are placed in the wrong slot (REI).
image

(meanwhile JEI shows no canning recipes when looking at uses for an empty can)

src/main/java/dev/galacticraft/mod/compat/rei/common/display/DefaultCanningDisplay.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    private static List<EntryIngredient> getInputs(ItemStack output) {
+        List<EntryIngredient> inputs = new ArrayList<>();
+        inputs.add(EntryIngredients.of(GCItems.EMPTY_CAN));
+        CannedFoodItem.getContents(output).forEach(stack -> inputs.add(EntryIngredients.of(stack)));
+        return inputs;
+    }
+
+    public List<InputIngredient<EntryStack<?>>> getInputIngredients() {
```
marcus8448: Currently this method is unused (in all our REI *Display classes) so the slot index offset is never applied. It looks like it's probably meant to override `getInputIngredients(AbstractContainerMenu, Player)`.

src/main/java/dev/galacticraft/mod/compat/jei/category/JEICanningProgressBar.java
```diff
@@ -0,0 +1,136 @@
...
+            ItemStack itemStack2 = this.getItemStack(itemStack);
+            if (itemStack2.isEmpty()) return;
+
+            graphics.renderItem(itemStack2, 0, 0);
+            graphics.renderItemDecorations(this.minecraft.font, itemStack2, 0, 0);
+        }
+
+        @Override
+        @SuppressWarnings("removal")
+        public List<Component> getTooltip(ItemStack itemStack, TooltipFlag tooltipFlag) {
```
marcus8448, line 115: Is there something blocking us from using the non-deprecated version? (non-blocking)

src/main/java/dev/galacticraft/mod/compat/jei/category/JEICanningProgressBar.java
```diff
@@ -0,0 +1,136 @@
...
+        this.progress = FoodCannerProgressAnimation.renderForRecipeViewer(graphics, xOffset, yOffset, this.progress + delta / 50.0F, this.showRow);
+    }
+
+    public CurrentCanIngredientRenderer getIngredientRenderer(ItemStack fallback) {
+        return new CurrentCanIngredientRenderer(this, fallback);
+    }
+
+    public class CurrentCanIngredientRenderer implements IIngredientRenderer<ItemStack>, IRecipeSlotRichTooltipCallback {
+        private final Minecraft minecraft = Minecraft.getInstance();
+        private final JEICanningProgressBar progressWidget;
```
marcus8448: This is an inner class that is both non-static and keeps a copy of its parent through a field, which seems redundant. Could you drop the field? (or make it a static class)

Same thing applies to the REI/EMI variants.

marcus8448 (changes requested):

src/main/java/dev/galacticraft/mod/compat/jei/category/JEICanningProgressBar.java
Roelymole: This is why, it also crashes the game if the list is empty...

`error: JEICanningProgressBar.CurrentCanIngredientRenderer 
is not abstract and does not override abstract method getTooltip(ItemStack,TooltipFlag) in IIngredientRenderer`

It's a bit of an oversight on JEI's side.

src/main/java/dev/galacticraft/mod/compat/jei/category/JEICanningProgressBar.java
Roelymole: I think the neatest solution is to use what I have just learnt is called a 'qualified this'.

e.g. `JEICanningProgressBar.this.getProgress()`

src/main/java/dev/galacticraft/mod/compat/rei/common/display/DefaultCanningDisplay.java
Roelymole: I haven't noticed any problems, so perhaps the default method is sufficient and these can simply be removed?

marcus8448 (approved):
Thanks!

PR #514, opened 2025-08-20
feat: falling meteors

danilwhale:
basically, port of GC4's falling meteors, but a bit more cleaner


src/main/java/dev/galacticraft/mod/client/render/entity/FallingMeteorRenderer.java
```diff
@@ -0,0 +1,63 @@
...
+        matrices.translate(-0.5, -0.25, -0.5);
+
+        this.dispatcher.renderBatched(
+                METEOR_BLOCK_STATE,
+                entity.blockPosition(),
+                entity.level(),
+                matrices,
+                vertexConsumers.getBuffer(ItemBlockRenderTypes.getMovingBlockRenderType(METEOR_BLOCK_STATE)),
+                false,
+                RandomSource.create() // this is probably going to blow up hard
```
Roelymole: You should be able to use `entity.getRandom()`

src/main/java/dev/galacticraft/mod/client/render/entity/FallingMeteorRenderer.java
```diff
@@ -0,0 +1,63 @@
...
+        );
+
+        matrices.popPose();
+    }
+
+    @Override
+    public ResourceLocation getTextureLocation(FallingMeteorEntity entityGoalInfo) {
```
Roelymole: Not that it particularly matters, but I can't quite understand where `entityGoalInfo` is coming from.

Just using `entity` like in the `render` method above should suffice.

src/main/java/dev/galacticraft/mod/content/entity/FallingMeteorEntity.java
```diff
@@ -0,0 +1,131 @@
...
+                    GCBlocks.FALLEN_METEOR.defaultBlockState().setValue(FallenMeteorBlock.HEAT, 5),
+                    Block.UPDATE_ALL);
+        }
+
+        this.discard();
+    }
+
+    private void onImpact(HitResult hitResult) {
+        if (hitResult.getType() == HitResult.Type.ENTITY) {
+            Entity entity = ((EntityHitResult) hitResult).getEntity();
```
Roelymole: This is a bit more succinct and should do the same thing.

```suggestion
        if (hitResult instanceof EntityHitResult entityHitResult) {
            Entity entity = entityHitResult.getEntity();
```

src/main/java/dev/galacticraft/mod/content/entity/FallingMeteorEntity.java
```diff
@@ -0,0 +1,131 @@
...
+                    this.getX(), this.getY() + 1.0 + this.random.nextDouble(), this.getZ() + this.random.nextDouble(),
+                    0.0, 0.0, 0.0);
+            this.level().addParticle(
+                    ParticleTypes.SMOKE,
+                    this.getX() - this.random.nextDouble() * 0.5, this.getY() + 1.0 + this.random.nextDouble(), this.getZ(),
+                    0.0, 0.0, 0.0);
+            this.level().addParticle(
+                    ParticleTypes.SMOKE,
+                    this.getX(), this.getY() + 1.0 + this.random.nextDouble(), this.getZ() - this.random.nextDouble(),
+                    0.0, 0.0, 0.0);
```
Roelymole: There is quite a lot of repetition that could be reduced. Probably the best way of doing this would be with a helper function, since I can't spot a straightforward pattern that would lend itself to a loop.

src/main/java/dev/galacticraft/mod/mixin/ServerPlayerMixin.java
```diff
@@ -124,4 +131,47 @@ private void checkIfCryoBedBlocked(BlockPos sleepingPos, Direction direction, Ca
...
+
+        // throw bigger meteor if you're lucky enough
+        if (this.level().random.nextInt(chance * 3) == 0) {
+            this.throwMeteor(6);
+        }
+    }
+
+    @Unique
+    private void throwMeteor(int size) {
+        // TODO: maybe move this into ServerLevel instead?
```
maxryan008: If you can I think making this a Server tick event would be neater then a mixin. Try doing at the end of a server tick have the server get all players and then spawn meteors based off random chance.

Roelymole:
This would currently fail our pull request checks since your two new classes do not have the license header and there are a few unused imports, however it is quite a simple thing to sort out just by running the `spotlessApply` gradle task.

e.g. `.\gradlew spotlessApply` or `.\gradlew.bat spotlessApply` for Windows.

src/main/java/dev/galacticraft/mod/content/entity/FallingMeteorEntity.java
```diff
@@ -0,0 +1,131 @@
...
+
+    public void setSize(int size) {
+        this.entityData.set(DATA_SIZE, size);
+    }
+
+    private void fallTo(Position position) {
+        this.level().explode(
+                this,
+                position.x(), position.y(), position.z(),
+                this.getSize() / 3 * 2,
```
Roelymole: I think swapping the order of `/ 3` and `* 2` would be clearer.

```suggestion
                this.getSize() * 2 / 3,
```

src/main/java/dev/galacticraft/mod/client/render/entity/FallingMeteorRenderer.java
```diff
@@ -0,0 +1,63 @@
...
+
+    public FallingMeteorRenderer(EntityRendererProvider.Context ctx) {
+        super(ctx);
+        this.shadowRadius = 1.0f;
+        this.dispatcher = ctx.getBlockRenderDispatcher();
+    }
+
+    @Override
+    public void render(FallingMeteorEntity entity, float yaw, float tickDelta, PoseStack matrices, MultiBufferSource vertexConsumers, int light) {
+        float size = entity.getSize();
```
Roelymole: Since you don't use this anywhere else, it might be more efficient to multiply by `0.5f` here (a single time), rather than later (3 times).

Roelymole (changes requested):
Here are a few things that stand out to me straight away.

src/main/java/dev/galacticraft/mod/client/render/entity/FallingMeteorRenderer.java
danilwhale: it was autogenerated by intellij, no idea where it gets `entityGoalInfo` from, but it's everywhere in my instance

src/main/java/dev/galacticraft/mod/content/entity/FallingMeteorEntity.java
Roelymole: That is so much more readable, thanks!

src/main/java/dev/galacticraft/mod/events/GCEventHandlers.java
```diff
@@ -75,6 +82,48 @@ public static void onWorldTick(ServerLevel level) {
...
+                throwMeteor(server, level, player, 1);
+            }
+
+            // throw bigger meteor if you're lucky enough
+            if (level.random.nextInt(chance * 3) == 0) {
+                throwMeteor(server, level, player, 6);
+            }
+        });
+    }
+    private static void throwMeteor(MinecraftServer server, ServerLevel level, Player targetPlayer, int meteorSize) {
```
Roelymole: There should be a blank line between these two methods.

src/main/java/dev/galacticraft/mod/events/GCEventHandlers.java
```diff
@@ -75,6 +82,48 @@ public static void onWorldTick(ServerLevel level) {
...
+        Player nearestPlayer = level.getNearestPlayer(targetPlayer, 100.0);
+        if (nearestPlayer == null || nearestPlayer.getId() > targetPlayer.getId()) return;
+
+        int maxOffset = server.getPlayerList().getViewDistance() * Constant.Chunk.WIDTH - 1;
+
+        int offsetX = Math.min(maxOffset, level.random.nextInt(20) + 160);
+        int offsetZ = level.random.nextInt(20) - 10;
+        double deltaX = level.random.nextDouble() * 2.0 - 2.5;
+        double deltaZ = level.random.nextDouble() * 5.0 - 2.5;
```
Roelymole: Is `deltaX` meant to be skewed towards negative values? The range of values `deltaX` can take is -4.5 to -0.5, since `nextDouble()` is between 0 and 1, which just seems a bit strange (although maybe there is a good reason I don't know about).

Doing something like this might be more intuitive:

```suggestion
        double deltaX = (level.random.nextDouble() - 0.5) * 2.0;
        double deltaZ = (level.random.nextDouble() - 0.5) * 5.0;
```

src/main/java/dev/galacticraft/mod/client/render/entity/FallingMeteorRenderer.java
```diff
@@ -0,0 +1,80 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+package dev.galacticraft.mod.client.render.entity;
+
+import com.mojang.blaze3d.vertex.PoseStack;
+import com.mojang.math.Axis;
+import dev.galacticraft.mod.Constant;
+import dev.galacticraft.mod.content.GCBlocks;
```
Roelymole: A static import of just `FALLEN_METEOR` might be better, since this is the only thing that is being used.

e.g. `import static dev.galacticraft.mod.content.GCBlocks.FALLEN_METEOR;`

The convention is to put static imports after all of the other imports with a blank line separating them.

src/main/java/dev/galacticraft/mod/client/render/entity/FallingMeteorRenderer.java
```diff
@@ -0,0 +1,80 @@
...
+import net.minecraft.client.renderer.block.BlockRenderDispatcher;
+import net.minecraft.client.renderer.entity.EntityRenderer;
+import net.minecraft.client.renderer.entity.EntityRendererProvider;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.level.block.state.BlockState;
+
+@Environment(EnvType.CLIENT)
+public class FallingMeteorRenderer extends EntityRenderer<FallingMeteorEntity> {
+    private static final ResourceLocation TEXTURE = Constant.id("textures/block/fallen_meteor.png");
+    private static final BlockState METEOR_BLOCK_STATE = GCBlocks.FALLEN_METEOR.defaultBlockState();
```
Roelymole: If you make the static import change, then the `GCBlocks.` part should be removed.

src/main/java/dev/galacticraft/mod/content/entity/FallingMeteorEntity.java
Roelymole, line 1: The comments about static imports apply to this class too.

src/main/java/dev/galacticraft/mod/events/GCEventHandlers.java
```diff
@@ -75,6 +82,48 @@ public static void onWorldTick(ServerLevel level) {
         level.galacticraft$getSealerManager().tick();
     }
 
+    public static void onServerTick(MinecraftServer server) {
+        // throw meteors around players
+        PlayerLookup.all(server).forEach(player -> {
+            ServerLevel level = player.serverLevel();
+            Holder<CelestialBody<?, ?>> celestialBody = level.galacticraft$getCelestialBody();
+            if (celestialBody == null) return;
```
Roelymole: We probably also want to return early if the celestial body is the Earth, since it might break the progression if meteoric iron is obtainable in the Overworld.

src/main/java/dev/galacticraft/mod/mixin/ServerPlayerMixin.java
Roelymole, line 1: You should run `spotlessApply` again to remove the unused imports in this file now that you are using the `END_SERVER_TICK` event.

src/main/java/dev/galacticraft/mod/content/GCEntityTypes.java
```diff
@@ -126,6 +126,13 @@ public class GCEntityTypes {
             .clientTrackingRange(8)
             .build());
 
+    public static final EntityType<FallingMeteorEntity> FALLING_METEOR = ENTITIES.register(Entity.FALLING_METEOR, EntityType.Builder.of(FallingMeteorEntity::new, MobCategory.AMBIENT)
+            .clientTrackingRange(150)
```
Roelymole: 150 seems very large considering this is the number of chunks.

danilwhale: what value should i use instead? 64?

Roelymole: The largest value used in Vanilla is 32 and that is for players. End Crystals, Lightning Bolts and Wardens have a range of 16, which is the next largest, so even that might be fine.

src/main/java/dev/galacticraft/mod/events/GCEventHandlers.java
```diff
@@ -94,9 +95,10 @@ public static void onServerTick(MinecraftServer server) {
             // calculate frequency of meteors on current celestial body
             float atmospherePressure = celestialBody.value().atmosphere().pressure();
             float frequency = (atmospherePressure <= Mth.EPSILON) ? 5.0f : (atmospherePressure * 100.0f);
+            frequency *= 1.0f / Galacticraft.CONFIG.meteorSpawnMultiplier();
```
Roelymole: Why don't you just use the `/=` assignment operator?

```suggestion
            frequency /= Galacticraft.CONFIG.meteorSpawnMultiplier();
```

Roelymole: Oh, you should also be careful about dividing by zero. You may want to deal with this by setting a minimum value in `ConfigImpl.java`.

src/main/java/dev/galacticraft/mod/events/GCEventHandlers.java
```diff
@@ -75,6 +85,50 @@ public static void onWorldTick(ServerLevel level) {
...
+            // throw meteor
+            int chance = Math.max(1, (int) (frequency * 750.0f));
+            if (level.random.nextInt(chance) == 0) {
+                throwMeteor(server, level, player, 1);
+            }
+
+            // throw bigger meteor if you're lucky enough
+            if (level.random.nextInt(chance * 3) == 0) {
+                throwMeteor(server, level, player, 6);
+            }
```
Roelymole, line 112: Technically these can both occur at the same time, which might not be what we want. Perhaps there should be a check to determine whether a meteor will be thrown and then a second check to determine which size the meteor should be.

danilwhale: this originates from gc4 source code, which had the same thing. i don't see any issue with spawning 2 meteors of different sizes at the same time. chance of spawning one meteor is already low, like 1 in 3750.

Roelymole: Sure, I guess my main thought would be that it reduces the number of random integers we need to generate, but the impact would be marginal.

Roelymole (approved):
Thanks for making all those changes!

src/main/java/dev/galacticraft/mod/events/GCEventHandlers.java
Roelymole: I have realised why `deltaX` is skewed towards negative values, it is because `offsetX` is always positive, so it sort of balances out. This system is a bit flawed, as the meteors can only come from the East and the distribution is not ideal. It might be best to revert this back to the previous expression until this can be redesigned properly, unless you want to have a go at that now.

Sorry for the confusion.

PR #516, opened 2025-08-22
feat: pre-commit and pre-push git hooks

Roelymole:
In order to enable the Git Hooks in this repository, which are useful to avoid lots commits to fix spotless violations, use the following command:
```shell
git config --local core.hooksPath .githooks/
```

The pre-commit Git Hook runs the `spotlessCheck` Gradle task before each git commit and aborts committing if the check fails, after running the `spotlessApply` Gradle task.

The pre-push Git Hook runs the `test` Gradle task before each git push and aborts pushing if the test fails.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -90,7 +90,7 @@ In order to enable the Git Hooks in this repository, use the following command:
 git config --local core.hooksPath .githooks/
 ```
 
-The pre-push Git Hook runs the spotlessCheck Gradle task before each Git Push and aborts pushing if the check fails. This is useful to avoid a lot of commits to fix spotless violations.
+The pre-commit Git Hook runs the spotlessCheck Gradle task before each git commit and aborts committing if the check fails. The pre-push Git Hook runs the spotlessCheck Gradle task before each git push and aborts pushing if the check fails. These are useful to avoid a lot of commits to fix spotless violations.
 
 #### Commit Message
 We use [conventional commits][conventional commits] specification to write commit messages.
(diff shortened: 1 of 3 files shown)


Roelymole:
The `test` Gradle task should probably only be run by the pre-push hook when pushing to TeamGalacticraft:main, since it does take a minute or so to complete and is covered by the Pull Request Check GitHub Action.

.github/CONTRIBUTING.md
```diff
@@ -79,14 +79,32 @@ If any files are missing the header, you can add it by executing the following G
 ./gradlew updateLicenses
```
marcus8448: While you're here, could you remove this since that task no longer exists?

.github/CONTRIBUTING.md
```diff
@@ -79,14 +79,32 @@ If any files are missing the header, you can add it by executing the following G
...
 However, we tend to squash-merge pull requests, so you are not required to do this.
 
 
 [actions]: https://github.com/TeamGalacticraft/Galacticraft/actions/workflows/build.yml?query=branch%3Amain+is%3Asuccess
 [conventional commits]: https://www.conventionalcommits.org
 [crowdin]: https://teamgalacticraft.crowdin.com/galacticraft
+[datagen]: /src/main/java/dev/galacticraft/mod/data
 [discord]: https://discord.gg/n3QqhMYyFK
 [generated]: /src/main/generated
 [license header]: /LICENSE_HEADER.txt
```
marcus8448: Also this file

.github/CONTRIBUTING.md
Roelymole: Sure, no problem.

PR #517, opened 2025-08-23
Oxygen 2

marcus8448:
Replaces the old state-inversion based oxygen system with a new "provider-based" system, and updates the oxygen sealer and oxygen bubble distributor implementations to match. This allows for precise point based oxygen checks rather than block position based checks.

Additionally adds the ability for oxygen sealers to quickly update their status based on broken/placed blocks:

- If a wall block is broken and causes a leak, the seal will break immediately
- If a wall block is broken and just exposes another sealable space, the seal will update to fill that space immediately
- If a block is placed in the middle of a sealed space, the seal will be maintained
- If a block is placed in a way that it blocks off an important/connecting corridor it will destroy the seal in the obstructed room while maintaining the seal on the other side.

Because of this, sealers now only recheck their seals when there is a block broken/placed, rather than on a set interval (the interval is only used for initial seals, to prevent constant rechecking of large unsealable areas).

User facing changes in action: https://www.youtube.com/watch?v=NoDF3qlwULY

## General Implementation Notes
Consider "breathable" to be equivalent  to "has atmosphere" or "has air" for simplicity, as we don't currently support anything more than that.

`AtmosphereProvider`: A block entity that has the ability to create breathable spaces. It keeps track of the precise places that it provides air for and can be queried.

A `Level` still keeps track of whether it has a breathable atmosphere, but now it won't bother checking for positional breathability if it does. This removes the ability to make "vacuums" (non-breathable spaces) on normally breathable planets, but that functionality wasn't in use anyways.

Should a `Level` not be breathable, then it queries the `Chunk` at the given x/z-coordinate, which then delegates to the `ChunkSection` at the given y-coordinate. The section returns a list of positions for `AtmosphereProvider`s which will then be queried (for the block entity). If any provider says that the position is breathable, it's considered breathable.

"Wall" and other solid blocks that are in or surround sealed spaces now are marked breathable.

## Unsolved bits

- [x] Does `AtmosphereProvider` lookup actually need to have per-block-resolution? **No, dropped.**
 - I would imagine most bases have only 1-2 sealers, and then maybe a couple bubble distributors?
   - Iterating over all providers in a section (at most maybe 4 items) and then doing a hash lookup for each one doesn't sound too bad in my head.
 - I'm leaning towards dropping it.
- [x] What should (not) be exposed in the public-facing api. **Public api moved to chunk level**
 - The section level interface must be used to attach a new `AtmosphereProvider` to a section
   - Too low level?
 - Chunk-level interface is almost empty (is there a use case?)
   - More complicated as most breathable checks require a world instance (as providers may be outside of the chunk)
-  [x] Multiple sealers can seal the same space **Fixed, with interface to identify.**
 - Could just add a simple check for other sealers, but no "nice" way to differentiate between bubble distributor (where overlap is ok) and sealer behaviour

## The future

There's a few neat things that could come out of this in the future, but aren't implemented here for simplicity:
- Variable air consumption based on blocks/entities in a bubble/sealed space (rather than flat volume-based rate)
- Air composition (we'll need to figure out a proper gas api first)

I've also been thinking about sealed spaces unattached to provider blocks (e.g. if a oxygen sealer is broken the air shouldn't just vanish), but that could get more complicated w.r.t block changes, composition/splitting.

### DIFF
--- a/src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenSealerBlockEntity.java
+++ b/src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenSealerBlockEntity.java
@@ -158,10 +158,6 @@ public class OxygenSealerBlockEntity extends MachineBlockEntity implements Space
         } else if (--this.sealCheckTime == 0) {
             this.sealCheckTime = SEAL_CHECK_TIME;
 
-            if (level.getBlockState(pos.relative(Direction.UP)).isCollisionShapeFullBlock(this.level, pos.relative(Direction.UP))) {
-                return GCMachineStatuses.BLOCKED;
-            }
-
             // CALCULATE SEALED SPACE:
             BlockPos.MutableBlockPos mutable = new BlockPos.MutableBlockPos();
             Object2BooleanOpenHashMap<BlockPos> visited = new Object2BooleanOpenHashMap<>();
@@ -361,6 +357,16 @@ public class OxygenSealerBlockEntity extends MachineBlockEntity implements Space
                 // UUUUU <- if we only checked WP, then newly disconnected pockets below the sealer would be considered connected
                 BlockPos target = this.worldPosition.relative(Direction.UP);
 
+                // special case: block placed on top of sealer
+                if (target.equals(pos)) {
+                    this.destroySeal();
+                    Object2BooleanOpenHashMap<BlockPos> sealed = new Object2BooleanOpenHashMap<>();
+                    sealed.put(pos, true);
+                    this.sealBlocks(sealed);
+                    this.markChanged();
+                    return;
+                }
+
                 // check that all surrounding blocks are still somehow connected to the sealed space.
                 boolean anyPass = false;
                 for (Direction direction : Direction.values()) {
@@ -377,6 +383,7 @@ public class OxygenSealerBlockEntity extends MachineBlockEntity implements Space
                 }
                 if (!anyPass) {
                     // This should not happen???
+                    Constant.LOGGER.warn("Block placement resulted in total seal destruction?");
                     this.destroySeal();
                 }
             }
@@ -421,6 +428,11 @@ public class OxygenSealerBlockEntity extends MachineBlockEntity implements Space
 
     // attempts to navigate (via BFS) to the target position
     private boolean tryNavigateTo(BlockPos.MutableBlockPos mutable, BlockPos target, ObjectOpenHashSet<BlockPos> visited, ObjectOpenHashSet<BlockPos> visitedNonSolid) {
+        if (this.sealedPositions.getOrDefault(mutable, false) || mutable.equals(target)) {
+            visited.add(mutable.immutable());
+            return mutable.equals(target);
+        }
+
         BlockPos start = mutable.immutable();
         ObjectArrayFIFOQueue<BlockPos> queue = new ObjectArrayFIFOQueue<>(32);
         queue.enqueue(start);


src/main/java/dev/galacticraft/impl/internal/mixin/oxygen/LevelChunkMixin.java
```diff
@@ -22,110 +22,58 @@
...
-    private void resetAirOnBlockChange(BlockPos pos, BlockState blockState, boolean bl, CallbackInfoReturnable<BlockState> cir) {
-        // TODO: Better system for checking whether the oxygen should be reset
-        if (blockState.isSolidRender(this.level, pos)) {
-            this.galacticraft$setInverted(pos.getX() & 15, pos.getY(), pos.getZ() & 15, false);
-        }
-    }
-
-    @WrapOperation(method = "setBlockState", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/level/block/state/BlockState;onPlace(Lnet/minecraft/world/level/Level;Lnet/minecraft/core/BlockPos;Lnet/minecraft/world/level/block/state/BlockState;Z)V", ordinal = 0))
-    private void extinguishFire(BlockState newState, Level level, BlockPos pos, BlockState oldState, boolean bl, Operation<Void> original) {
-        if (level.isBreathable(pos) || !GCEventHandlers.extinguishBlock(level, pos, newState)) {
```
Roelymole, line 127: You aren't calling `GCEventHandlers.extinguishBlock(level, pos, newState)` anywhere, so the extinguishable blocks are no longer being extinguished.

marcus8448: Oops - totally missed that.

This also reminds me that there's also no mechanism to alert blocks that the atmosphere has changed...

maxryan008:
Not sure if this was a feature already but does copper still oxidize in no atmosphere environments. specifically could you design it so a sealed room will let copper oxidize but in no atmosphere environments it will no longer oxidize

src/main/java/dev/galacticraft/api/accessor/GCBlockExtensions.java
```diff
@@ -0,0 +1,85 @@
...
+     * @param pos the position of the block in-world
+     * @param state the state being placed
+     * @return the transformed block state.
+     */
+    default BlockState galacticraft$extinguishBlockPlace(BlockPos pos, BlockState state) {
+        return state;
+    }
+
+    /**
+     * {@return whether the given block state needs to be altered to changes in atmospheric composition}
```
Roelymole: Do you mean alerted, not altered?

src/main/java/dev/galacticraft/api/accessor/LevelOxygenAccessorRO.java
```diff
@@ -0,0 +1,87 @@
...
+
+package dev.galacticraft.api.accessor;
+
+import dev.galacticraft.api.block.entity.AtmosphereProvider;
+import net.minecraft.core.BlockPos;
+import net.minecraft.world.phys.Vec3;
+
+import java.util.Iterator;
+
+public interface LevelOxygenAccessorRO {
```
Roelymole, line 31: What does the RO stand for, out of curiosity?

src/main/java/dev/galacticraft/api/accessor/ChunkOxygenAccessor.java
Roelymole, line 1: Is there a difference between atmosphere provider and atmospheric provider?

You have used both in this file (and elsewhere).

src/main/java/dev/galacticraft/impl/internal/mixin/oxygen/ChunkSerializerMixin.java
```diff
@@ -22,53 +22,36 @@
...
-        if (!accessor.galacticraft$isEmpty()) {
-            CompoundTag nbt = new CompoundTag();
-            nbt.putByteArray(Constant.Nbt.OXYGEN, Objects.requireNonNull(accessor.galacticraft$getBits()).toByteArray());
-            nbtCompound2.put(Constant.Nbt.GC_API, nbt);
-        }
+    @Inject(method = "write", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/level/chunk/LevelChunkSection;getStates()Lnet/minecraft/world/level/chunk/PalettedContainer;"))
+    private static void serializeOxygen(ServerLevel world, ChunkAccess chunk, CallbackInfoReturnable<CompoundTag> cir, @Local(ordinal = 1) CompoundTag sectionTag, @Local LevelChunkSection section) {
+        ChunkSectionOxygenAccessor accessor = (ChunkSectionOxygenAccessor) section;
+        CompoundTag apiTag = new CompoundTag();
+        accessor.galacticraft$writeTag(apiTag);
```
Roelymole: Since `accessor` is only used once, this might be fine.

```suggestion
        CompoundTag apiTag = new CompoundTag();
        ((ChunkSectionOxygenAccessor) section).galacticraft$writeTag(apiTag);
```

src/main/java/dev/galacticraft/impl/internal/mixin/oxygen/ChunkSerializerMixin.java
```diff
@@ -22,53 +22,36 @@
...
-        CompoundTag apiCompound = compoundTag.getCompound(Constant.Nbt.GC_API);
-        if (apiCompound.contains(Constant.Nbt.OXYGEN, Tag.TAG_BYTE_ARRAY)) {
-            ((ChunkSectionOxygenAccessor) levelChunkSection).galacticraft$setBits(BitSet.valueOf(apiCompound.getByteArray(Constant.Nbt.OXYGEN)));
-        } else {
-            ((ChunkSectionOxygenAccessor) levelChunkSection).galacticraft$setBits(null);
-        }
+    @Inject(method = "read", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/entity/ai/village/poi/PoiManager;checkConsistencyWithBlocks(Lnet/minecraft/core/SectionPos;Lnet/minecraft/world/level/chunk/LevelChunkSection;)V"))
+    private static void deserializeOxygen(ServerLevel world, PoiManager poiStorage, RegionStorageInfo key, ChunkPos chunkPos, CompoundTag nbt, CallbackInfoReturnable<ProtoChunk> cir, @Local(ordinal = 1) CompoundTag sectionTag, @Local LevelChunkSection levelChunkSection) {
+        ((ChunkSectionOxygenAccessor) levelChunkSection).galacticraft$readTag(sectionTag.getCompound(Constant.Nbt.GC_API));
```
Roelymole: It might be nicer to shorten `levelChunkSection` to just `section` like in `serializeOxygen` above (or at least make them consistent).

src/main/java/dev/galacticraft/impl/internal/mixin/oxygen/extinguish/AbstractCandleBlockMixin.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @Override
+    public void galacticraft$onAtmosphereChange(ServerLevel level, BlockPos pos, BlockState state, boolean breathable) {
+        if (!breathable) {
+            this.getParticleOffsets(state).forEach(vec3 ->
+                    level.sendParticles(ParticleTypes.SMOKE, (double) pos.getX() + vec3.x(), (double) pos.getY() + vec3.y(), (double) pos.getZ() + vec3.z(), 0, 0.0D, 0.1D, 0.0D, 0.0D)
+            );
+            level.playSound(null, pos, SoundEvents.CANDLE_EXTINGUISH, SoundSource.BLOCKS, 1.0F, 1.0F);
+
+            level.setBlock(pos, state.setValue(LIT, false), 11);
```
Roelymole: Using `Block.UPDATE_ALL_IMMEDIATE` would be more readable than using the value of 11 directly.

src/main/java/dev/galacticraft/impl/internal/mixin/oxygen/extinguish/CampfireBlockMixin.java
```diff
@@ -0,0 +1,80 @@
...
+        return state.getValue(LIT);
+    }
+
+    @Override
+    public void galacticraft$onAtmosphereChange(ServerLevel level, BlockPos pos, BlockState state, boolean breathable) {
+        if (!breathable) {
+            // todo: assuming all campfire = actual fire? should there be smoke?
+            level.playSound(null, pos, SoundEvents.GENERIC_EXTINGUISH_FIRE, SoundSource.BLOCKS, 0.2F, 1.0F);
+
+            level.setBlock(pos, state.setValue(CampfireBlock.LIT, false), 11);
```
Roelymole: Again, using `Block.UPDATE_ALL_IMMEDIATE` would be more readable than using the value of 11 directly.

src/main/java/dev/galacticraft/impl/internal/mixin/oxygen/extinguish/LanternBlockMixin.java
```diff
@@ -0,0 +1,98 @@
...
+import net.minecraft.world.item.context.BlockPlaceContext;
+import net.minecraft.world.level.block.Block;
+import net.minecraft.world.level.block.Blocks;
+import net.minecraft.world.level.block.LanternBlock;
+import net.minecraft.world.level.block.state.BlockState;
+import org.spongepowered.asm.mixin.Mixin;
+
+@Mixin(LanternBlock.class)
+public class LanternBlockMixin implements GCBlockExtensions {
+    @WrapMethod(method = "getStateForPlacement")
```
Roelymole: `@ModifyReturnValue` would be a more appropriate choice, rather than using `@WrapMethod` and unconditionally calling the original method.

https://github.com/LlamaLad7/MixinExtras/wiki/ModifyReturnValue

src/main/java/dev/galacticraft/impl/internal/mixin/oxygen/extinguish/SoulFireBlockMixin.java
```diff
@@ -0,0 +1,68 @@
...
+import net.minecraft.world.level.LevelReader;
+import net.minecraft.world.level.block.Block;
+import net.minecraft.world.level.block.Blocks;
+import net.minecraft.world.level.block.SoulFireBlock;
+import net.minecraft.world.level.block.state.BlockState;
+import org.spongepowered.asm.mixin.Mixin;
+
+@Mixin(SoulFireBlock.class)
+public class SoulFireBlockMixin implements GCBlockExtensions {
+    @WrapMethod(method = "canSurvive")
```
Roelymole: I think using `@ModifyReturnValue` would be more suitable here too.

src/main/java/dev/galacticraft/impl/internal/mixin/oxygen/extinguish/FireBlockMixin.java
```diff
@@ -0,0 +1,51 @@
...
+    public boolean galacticraft$hasLegacyExtinguishTransform(BlockState state) {
+        return true;
+    }
+
+    @Override
+    public BlockState galacticraft$extinguishBlockPlace(BlockPos pos, BlockState state) {
+        return Blocks.AIR.defaultBlockState();
+    }
+
+    @WrapMethod(method = "canSurvive")
```
Roelymole: `@ModifyReturnValue` could be used here instead too.

src/main/java/dev/galacticraft/impl/internal/mixin/oxygen/extinguish/CampfireBlockMixin.java
```diff
@@ -0,0 +1,80 @@
...
+
+    @Override
+    public boolean galacticraft$hasAtmosphereListener(BlockState state) {
+        return state.getValue(LIT);
+    }
+
+    @Override
+    public void galacticraft$onAtmosphereChange(ServerLevel level, BlockPos pos, BlockState state, boolean breathable) {
+        if (!breathable) {
+            // todo: assuming all campfire = actual fire? should there be smoke?
```
Roelymole: I hadn't bothered before because they are constantly producing smoke anyway.

src/main/java/dev/galacticraft/impl/internal/mixin/oxygen/LevelChunkSectionMixin.java
```diff
@@ -22,98 +22,148 @@
...
-    public void galacticraft$setInverted(int pos, boolean value) {
-        if (value) {
-            if (this.bits == null) this.bits = new BitSet(pos);
-            this.bits.set(pos);
-        } else if (this.bits != null) {
-            this.bits.clear(pos);
+    public boolean galacticraft$addProvider(BlockPos pos) {
+        if (this.providers == null) {
+            this.providers = new ArrayList<>(1);
+            this.providers.add(pos);
```
Roelymole: You may as well use `Lists.newArrayList` here too, right?

```suggestion
            this.providers = Lists.newArrayList(pos);
```

src/main/java/dev/galacticraft/impl/internal/mixin/oxygen/LevelChunkSectionMixin.java
```diff
@@ -22,98 +22,148 @@
...
+            return original.call() + VarInt.getByteSize(this.providers.size()) + Long.BYTES * this.providers.size();
+        } else {
+            return original.call() + VarInt.getByteSize(0);
+        }
+    }
+
+    @WrapMethod(method = "hasOnlyAir()Z")
+    private boolean verifyOxygenEmpty(Operation<Boolean> original) {
+        return original.call() && this.galacticraft$isEmpty();
+    }
```
Roelymole: Both of these could use `@ModifyReturnValue` instead.

src/main/java/dev/galacticraft/impl/internal/oxygen/SortedPosList.java
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    public double getDistance(int index) {
+        return this.distances[index];
+    }
+
+    public boolean remove(BlockPos pos) {
+        return this.remove(pos, this.calculateDistanceSq(pos));
+    }
+    public boolean remove(BlockPos pos, double distance) {
```
Roelymole, line 106: Maybe this is where the extra blank line at the end of the `add(BlockPos pos, double distance)` method came from.
😜

src/main/java/dev/galacticraft/impl/internal/mixin/oxygen/extinguish/TorchBlockMixin.java
```diff
@@ -0,0 +1,71 @@
...
+@Mixin(TorchBlock.class)
+public class TorchBlockMixin implements GCBlockExtensions {
+    @Override
+    public boolean galacticraft$hasLegacyExtinguishTransform(BlockState state) {
+        return true;
+    }
+
+    @Override
+    public BlockState galacticraft$extinguishBlockPlace(BlockPos pos, BlockState state) {
+        return state.getBlock() == Blocks.TORCH ? GCBlocks.UNLIT_TORCH.withPropertiesOf(state) : state;
```
Roelymole: What about soul torches?

src/main/java/dev/galacticraft/impl/internal/mixin/oxygen/extinguish/WallTorchBlockMixin.java
```diff
@@ -0,0 +1,72 @@
...
+@Mixin(WallTorchBlock.class)
+public class WallTorchBlockMixin implements GCBlockExtensions {
+    @Override
+    public boolean galacticraft$hasLegacyExtinguishTransform(BlockState state) {
+        return true;
+    }
+
+    @Override
+    public BlockState galacticraft$extinguishBlockPlace(BlockPos pos, BlockState state) {
+        return state.getBlock() == Blocks.WALL_TORCH ? GCBlocks.UNLIT_WALL_TORCH.withPropertiesOf(state) : state;
```
Roelymole: What about soul torches?

src/main/java/dev/galacticraft/api/accessor/ChunkOxygenAccessor.java
marcus8448: Nope - just inconsistent naming. I'll update it to the former.

src/main/java/dev/galacticraft/api/accessor/LevelOxygenAccessorRO.java
marcus8448: It stands for read-only. The interface is split so that `LevelReader` can implement the read-only parts. It could be renamed to `LevelOxygenReader` or something if that's easier to understand.

Roelymole: Ah, that makes sense. Don't feel you have to rename it, but a comment saying read-only might be nice.

src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenSealerBlockEntity.java
Roelymole, line 1: It seems like `MAX_SEALER_VOLUME` is being used as the maximum number of blocks that can be sealed, regardless of how many sealers there are. I feel like this probably shouldn't be the case.

Roelymole: It also seems to be the case that covering up an oxygen sealer/subdividing a sealed room does not update the space without a sealer anymore to be non-breathable.

https://github.com/user-attachments/assets/3d96d300-7898-4164-a3cd-166741b9ceaf

PR #518, opened 2025-09-06
feat: more specialized REI recipe viewers (with animations)

Roelymole:
- Split the compressing recipe viewer into two: one for the coal-powered compressor and one for the electric compressor.
This allows for different processing times to be displayed and also the ability for the electric compressor to process two instances of a recipe at the same time.

https://github.com/user-attachments/assets/f650cf2b-b689-4ce4-bc71-27a2d2dd99d6

- Similarly, there are separate recipe viewers for the electric furnace and the electric arc furnace (previously using the smelting/blasting recipe viewers respectively).
The output for the electric arc furnace swaps between one and two items to indicate that there is a chance of bonus items.

https://github.com/user-attachments/assets/82a3071d-1b1b-4d78-93df-7ecb8776a0b4

- Also animated the progress arrows.

### To Do:
- [x] <del>Animate the progress indicator in the circuit fabricator recipe viewer.</del> *Completed in commit d478db09d2c909f7f5f50982a9c96ba3839bc383*
- [x] Improve the look of the electric (arc) furnace displays

### DIFF
--- a/src/main/java/dev/galacticraft/mod/compat/rei/client/category/CustomArrowWidget.java
+++ b/src/main/java/dev/galacticraft/mod/compat/rei/client/category/CustomArrowWidget.java
@@ -23,7 +23,7 @@
 package dev.galacticraft.mod.compat.rei.client.category;
 
 import me.shedaniel.math.Rectangle;
-import me.shedaniel.rei.api.client.gui.widgets.Arrow;
+import me.shedaniel.rei.api.client.gui.widgets.WidgetWithBounds;
 import net.minecraft.client.gui.GuiGraphics;
 import net.minecraft.client.gui.components.events.GuiEventListener;
 import net.minecraft.resources.ResourceLocation;
@@ -32,7 +32,7 @@ import net.minecraft.util.Mth;
 import java.util.Collections;
 import java.util.List;
 
-public class CustomArrowWidget extends Arrow {
+public class CustomArrowWidget extends WidgetWithBounds {
     private final ResourceLocation texture;
     private Rectangle bounds;
     private int progressU;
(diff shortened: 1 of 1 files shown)


src/main/java/dev/galacticraft/mod/compat/rei/client/category/CustomArrowWidget.java
```diff
@@ -0,0 +1,81 @@
...
+import me.shedaniel.rei.api.client.gui.widgets.Arrow;
+import net.minecraft.client.gui.GuiGraphics;
+import net.minecraft.client.gui.components.events.GuiEventListener;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.util.Mth;
+
+import java.util.Collections;
+import java.util.List;
+
+public class CustomArrowWidget extends Arrow {
```
marcus8448: I think this would be better off just extending `WidgetWithBounds`, since the arrow-specific methods aren't really used (unless you have future plans for them). Otherwise it's a kind of an awkward constructor/builder pattern mix.

marcus8448: I think this would be better off just extending `WidgetWithBounds`, since the arrow-specific methods aren't really used (unless you have future plans for them). Otherwise it's a kind of an awkward constructor/builder pattern mix.

marcus8448 (commented):
Looks good! Just one small thing.

src/main/java/dev/galacticraft/mod/compat/rei/client/category/CustomArrowWidget.java
Roelymole: If I had any future plans I've forgotten them...

PR #520, opened 2025-09-16
Feature/api split

AlphaMode:
Reintroduces the API has a separate source set of the mod similar to how the old [GC-API](https://github.com/TeamGalacticraft/GalacticraftAPI) worked.
Currently the API doesn't have its own mod id/fmj.
However this may be changed depending on future API plans.
My main reason for creating this is to prevent accidental GC Mod code in the API, and also allow addons to use the api without depending on GC itself


src/api/java/dev/galacticraft/api/rocket/part/EmptyRocketLayout.java
Roelymole, line 1: This looks like it is still WIP, so maybe it should be saved for a future PR.

src/api/java/dev/galacticraft/api/rocket/part/GenericRocketLayout.java
Roelymole, line 1: This should probably be removed from this PR and maybe included in a separate PR later on.

src/main/java/dev/galacticraft/mod/client/render/entity/rocket/RocketEntityRenderer.java
```diff
@@ -81,6 +81,13 @@ public void render(RocketEntity entity, float yaw, float partialTick, PoseStack
 //        RenderSystem.setShaderTexture(0, getTextureLocation(entity));
         matrices.translate(0.0F, 0.4375F, 0.0F);
 
+//        RocketLayout layout = entity.getRocketLayout();
+//        List<Holder<? extends RocketPart<?, ?>>> parts = layout.parts().stream().map(eitherHolder -> maybeGet(entity, eitherHolder));
+//
+//        for (EitherHolder<? extends RocketPart<?, ?>> part : layout.parts()) {
+//            part.
+//        }
+
```
Roelymole, line 90: Is this still needed?

src/main/java/dev/galacticraft/mod/content/entity/vehicle/RocketEntity.java
```diff
@@ -701,6 +700,11 @@ public int getPreLaunchWait() {
         return holder.flatMap(tEitherHolder -> tEitherHolder.unwrap(this.registryAccess())).orElse(null);
     }
 
+//    @Override
+//    public @NotNull RocketLayout getRocketLayout() {
+//        return getRocketData().layout();
+//    }
+
```
Roelymole, line 712: Is this still needed?

PR #526, opened 2025-10-08
fix : make vacuum glass look acceptable

lingyun-create:


DarianLStephens:
To be fair, Minecraft doesn't have tilted hitboxes, either.
You have signs which can be rotated, but the hitbox is the same pillar.
Even lecterns just have a stair-step hitbox to vaguely match the slanted book receptacle.

PR #530, opened 2025-11-01
feat: implement the getCloneItemStack method for CannedFoodBlock

Roelymole:
Allow pick block (middle click) to be used to quickly get an individual canned food item from a stack of cans.

https://github.com/user-attachments/assets/08a6129c-6ba4-4b0a-9ce2-35a6bf3e7efd


Roelymole:
Rewatching the video, I have just noticed that the WTHIT/WAILA tooltip displays the item that would be obtained by using pick block as well!

marcus8448 (approved):
Cool!

Indexing `POSITIONS` based on can count looks scary but that's not a problem caused by this PR.

PR #538, opened 2025-12-27
feat: automatically set required Minecraft and Fabric API versions in fabric.mod.json

JustAPotota:
The game now shows an error message when trying to run Galacticraft without Fabric API installed or with the wrong Minecraft version:
image

### DIFF
--- a/src/main/resources/fabric.mod.json
+++ b/src/main/resources/fabric.mod.json
@@ -326,7 +326,7 @@
     }
   },
   "depends": {
-    "minecraft": ">=1.21",
+    "minecraft": "=$minecraftVersion",
     "fabricloader": ">=0.15.3",
     "fabric-api": ">=$fabricVersion"
   },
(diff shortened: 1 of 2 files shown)


src/main/resources/fabric.mod.json
```diff
@@ -327,7 +327,8 @@
   },
   "depends": {
     "minecraft": ">=1.21",
```
Roelymole: It might be good to do a similar thing for the Minecraft version here, since it is currently outdated and this approach should prevent that from happening again.

JustAPotota: Good idea. Can I also change it to `=$minecraftVersion`? I don't think a big content mod like GC5 will work on multiple versions at once, and mods with a `>=` always lead to headaches when a user decides to update their game.

Roelymole: I think we are probably quite safe to specify the Minecraft version exactly.

Roelymole (approved):

PR #539, opened 2025-12-28
fix: readded viscosity to galacticraft fluids

SureShotTime:
Fix Issue #301 that re-adds the basic vanilla water physics to Galacticraft fluids (sulfuric acid, fuel, and crude oil) providing GC4 functionality.

Updated naming convention

Can extend to custom viscosity and particles, but basic usability should be in the pre-alpha release.

### DIFF
--- a/src/main/generated/data/minecraft/tags/fluid/water.json
+++ b/src/main/generated/data/minecraft/tags/fluid/water.json
@@ -1,0 +1,7 @@
+{
+  "values": [
+    "#galacticraft:sulfuric_acid",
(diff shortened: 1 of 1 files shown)


Roelymole (commented):
If this is the approach we are going to use to address this issue then it would be better to generate this file using the datagen system (see the [GCFluidTagProvider](https://github.com/TeamGalacticraft/Galacticraft/blob/main/src/main/java/dev/galacticraft/mod/data/tag/GCFluidTagProvider.java) class).

PR #540, opened 2026-01-04
feature: Creative galacticraft gear inventory

MaverXka:
image

### DIFF
--- a/src/main/java/dev/galacticraft/mod/client/gui/widget/RadioButton.java
+++ b/src/main/java/dev/galacticraft/mod/client/gui/widget/RadioButton.java
@@ -74,15 +74,13 @@ public class RadioButton extends AbstractWidget {
     }
 
     public int getButtonHovered(int mouseX, int mouseY) {
-        boolean top = mouseX >= getX() && mouseX < getX() + BTN_WIDTH &&
-                mouseY >= getY() && mouseY < getY() + BTN_HEIGHT;
-
-        boolean bottom = mouseX >= getX() && mouseX < getX() + BTN_WIDTH &&
-                mouseY >= getY() + BTN_HEIGHT && mouseY < getY() + BTN_HEIGHT * 2;
-        if (top) {
+        if (mouseX >= getX() && mouseX < getX() + BTN_WIDTH &&
+                mouseY >= getY() && mouseY < getY() + BTN_HEIGHT) {
             return 0;
         }
-        if (bottom) {
+
+        if (mouseX >= getX() && mouseX < getX() + BTN_WIDTH &&
+                mouseY >= getY() + BTN_HEIGHT && mouseY < getY() + BTN_HEIGHT * 2) {
             return 1;
         }
         return -1;
--- a/src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
+++ b/src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
@@ -32,7 +32,6 @@ import dev.galacticraft.mod.screen.slot.AccessorySlot;
 import net.fabricmc.fabric.api.client.networking.v1.ClientPlayNetworking;
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.gui.GuiGraphics;
-import net.minecraft.client.gui.components.EditBox;
 import net.minecraft.client.gui.screens.Screen;
 import net.minecraft.client.gui.screens.inventory.CreativeModeInventoryScreen;
 import net.minecraft.client.gui.screens.inventory.EffectRenderingInventoryScreen;
@@ -68,15 +67,6 @@ public abstract class CreativeScreenMixin extends EffectRenderingInventoryScreen
 
     @Shadow protected abstract void selectTab(CreativeModeTab group);
 
-    @Shadow protected abstract void renderTabButton(GuiGraphics graphics, CreativeModeTab group);
-
-    @Shadow private EditBox searchBox;
-
-    @Shadow protected abstract boolean canScroll();
-
-    @Shadow @Final private static ResourceLocation SCROLLER_SPRITE;
-    @Shadow @Final private static ResourceLocation SCROLLER_DISABLED_SPRITE;
-    @Shadow private float scrollOffs;
     @Shadow @Nullable private Slot destroyItemSlot;
     @Shadow @Final private static SimpleContainer CONTAINER;
     @Unique private final List<AccessorySlot> gc$slots = new ArrayList<>();
@@ -118,6 +108,7 @@ public abstract class CreativeScreenMixin extends EffectRenderingInventoryScreen
     private ResourceLocation gc$changeBackground(ResourceLocation original) {
         return isGCInventoryEnabled() ? GC_GUIBG_TEX : original;
     }
+
     @ModifyArg(method = "renderBg", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/gui/screens/inventory/InventoryScreen;renderEntityInInventoryFollowsMouse(Lnet/minecraft/client/gui/GuiGraphics;IIIIIFFFLnet/minecraft/world/entity/LivingEntity;)V"), index = 1)
     private int gc$shiftEntityLeft(int original) {
         return isGCInventoryEnabled() ? leftPos + 55 : original;


src/main/generated/assets/galacticraft/lang/en_us.json
Roelymole, line 1: You shouldn't modify the generated files directly, since any changes will get overwritten when the runDatagen Gradle task is executed (using the command `./gradlew runDatagen`).

You should add these translations to the `generateConfigTranslations` method in the `GCTranslationProvider` class [here](https://github.com/TeamGalacticraft/Galacticraft/blob/c06cdaa718000ea6c638da51505453ae1ee118e5/src/main/java/dev/galacticraft/mod/data/GCTranslationProvider.java#L974).

src/main/java/dev/galacticraft/mod/accessor/GCCreativeGuiSlots.java
Roelymole, line 1: None of the brand new files you have created have the licence header at the top. Running the command `./gradlew spotlessApply` in the terminal should automatically add the license headers for you (as well as fixing a few other things such as whitespace issues).

src/main/java/dev/galacticraft/mod/client/gui/widget/RadioButton.java
```diff
@@ -0,0 +1,129 @@
...
+
+    private boolean isBottomButtonActive = false;
+
+    public boolean getIsBottomButtonActive() { return isBottomButtonActive; }
+
+    public Runnable radioButtonOnClick;
+
+    public RadioButton(int x, int y){
+        super(x,y,BTN_WIDTH, BTN_HEIGHT*2, Component.empty());
+    }
```
Roelymole:
```suggestion
    public RadioButton(int x, int y) {
        super(x, y, BTN_WIDTH, BTN_HEIGHT * 2, Component.empty());
    }
```

src/main/java/dev/galacticraft/mod/client/gui/widget/RadioButton.java
```diff
@@ -0,0 +1,129 @@
...
+public class RadioButton extends AbstractWidget {
+
+    ResourceLocation buttonTex = Constant.id("textures/gui/radiobutton_gear_inventory_buttons.png");
+
+    private static final int BTN_WIDTH = 11;
+    private static final int BTN_HEIGHT = 10;
+
+    private boolean isBottomButtonActive = false;
+
+    public boolean getIsBottomButtonActive() { return isBottomButtonActive; }
```
Roelymole: In most cases, I don't think much is gained by putting the body of methods like this on the same line.

```suggestion
    public boolean getIsBottomButtonActive() {
        return isBottomButtonActive;
    }
```

src/main/java/dev/galacticraft/mod/client/gui/widget/RadioButton.java
```diff
@@ -0,0 +1,129 @@
...
+                    {
+                        isBottomButtonActive = !isBottomButtonActive;
+                        this.playDownSound(Minecraft.getInstance().getSoundManager());
+                        this.onClick(mouseX, mouseY);
+                        radioButtonOnClick.run();
+                        return true;
+                    }
+                }
+
+
```
Roelymole: The code inside these if/else blocks is identical, so it would be better to avoid repeating it by combining the logic together inside the condition for the if statement.

```suggestion
                if (!isBottomButtonActive ? isBottomHovered : isTopHovered) {
                    isBottomButtonActive = !isBottomButtonActive;
                    this.playDownSound(Minecraft.getInstance().getSoundManager());
                    this.onClick(mouseX, mouseY);
                    radioButtonOnClick.run();
                    return true;
                }
```

src/main/java/dev/galacticraft/mod/client/gui/widget/RadioButton.java
```diff
@@ -0,0 +1,129 @@
...
+    // style - 0: default 1: active 2: inactive
+    private int getArrowBlitCoordsU(int style)
+    {
+        return switch (style) {
+            case 0 -> 0;
+            case 1 -> 0;
+            case 2 -> 11;
+            default -> 0;
+        };
```
Roelymole: The same thing can be achieved on a single line using a ternary operator.

```suggestion
        return style == 2 ? 11 : 0;
```

src/main/java/dev/galacticraft/mod/client/gui/widget/RadioButton.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    public int getButtonHovered(int MouseX, int MouseY)
+    {
```
Roelymole: Variables in Java should be in camelCase, not PascalCase, since PascalCase is used for class names. Not to mention this is inconsistent with other places where you haven't capitalised the first letter of `mouseX` and `mouseY`.

```suggestion
    public int getButtonHovered(int mouseX, int mouseY) {
```

src/main/java/dev/galacticraft/mod/client/gui/widget/RadioButton.java
```diff
@@ -0,0 +1,129 @@
...
+
+    // style - 0: default 1: active 2: inactive
+    private int getArrowBlitCoordsV(int style,boolean isBottomTex)
+    {
+        return switch (style) {
+            case 0 -> isBottomTex ? 226 : 216;
+            case 1 -> isBottomTex ? 246 : 236;
+            case 2 -> isBottomTex ? 226 : 216;
+            default -> isBottomTex ? 226 : 216;
+        };
```
Roelymole: This switch case statement is quite redundant since case 0, case 2 and default are identical.

src/main/java/dev/galacticraft/mod/client/gui/widget/RadioButton.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    @Override
+    protected void renderWidget(GuiGraphics context, int mouseX, int mouseY, float delta) {
+
+        int hovered = getButtonHovered(mouseX, mouseY);
+        boolean isTopHovered = hovered == 0;
+        boolean isBottomHovered = hovered == 1;
+        context.blit(buttonTex, getX(), getY(),getArrowBlitCoordsU(isBottomButtonActive ? (isTopHovered ? 1 : 0) : 2),getArrowBlitCoordsV(isBottomButtonActive ? (isTopHovered ? 1 : 0) : 2, false),BTN_WIDTH,BTN_HEIGHT,256,256);
+        context.blit(buttonTex, getX(), getY()+BTN_HEIGHT,getArrowBlitCoordsU(isBottomButtonActive ? 2 : (isBottomHovered ? 1 : 0)),getArrowBlitCoordsV(isBottomButtonActive ? 2 : (isBottomHovered ? 1 : 0), true),BTN_WIDTH,BTN_HEIGHT,256,256);
```
Roelymole: Please could you put a space after commas and preferably around operators such as the `+` here too.

src/main/java/dev/galacticraft/mod/mixin/AbstractContainerMenuAccessor.java
Roelymole, line 1: I'm guessing you had added some code to this file at some point and later deleted it.

One of the other things that running `./gradlew spotlessApply` does is remove unused imports, but for the sake of the git history it would be best to remove all of the changes you have made to this file.

src/main/java/dev/galacticraft/mod/screen/GCPlayerInventoryMenu.java
Roelymole, line 1: Again, for the sake of the git history it would be best to remove all of the changes you have made to this file.

src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
```diff
@@ -0,0 +1,489 @@
...
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+
+import java.lang.reflect.Constructor;
+import java.util.*;
+
+@Mixin(CreativeModeInventoryScreen.class)
+public abstract class CreativeScreenMixin implements GCInventoryFlag, GCCreativeGuiSlots {
```
Roelymole: You should be able to get rid of most of the accessors/invokers in `AbstractContainerScreenAccessor` and the entirety of `ScreenAccessor` by making this class extend `EffectRenderingInventoryScreen` (which in turn extends `AbstractContainerScreen` and `Screen`).

This will require a trivial constructor method such as the following (this isn't used by anything, it is just needed to be able to compile).

```suggestion
public abstract class CreativeScreenMixin extends EffectRenderingInventoryScreen implements GCInventoryFlag, GCCreativeGuiSlots {
    CreativeScreenMixin() {
        super(null, null, null);
    }
```

src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
```diff
@@ -0,0 +1,489 @@
...
+     */
+    @Overwrite
+    public void renderBg(GuiGraphics graphics, float delta, int mouseX, int mouseY)
+    {
+
+        CreativeModeInventoryScreen self = (CreativeModeInventoryScreen)(Object)this;
+        int leftPos = ((AbstractContainerScreenAccessor)(Object)this).getLeftPos();
+        int topPos = ((AbstractContainerScreenAccessor)(Object)this).getTopPos();
+        int imageWidth = ((AbstractContainerScreenAccessor)(Object)this).getImageWidth();
+        int imageHeight = ((AbstractContainerScreenAccessor)(Object)this).getImageHeight();
```
Roelymole: If this class extends `AbstractContainerScreen`, then all of this can be deleted as `this.leftPos`, `this.topPos`, `this.imageWidth` and `this.imageHeight` are available via inheritance.

src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
```diff
@@ -0,0 +1,489 @@
...
+        }
+
+    }
+
+    @Inject(method = "mouseClicked", at = @At("HEAD"), cancellable = true)
+    private void onMouseClicked(double mouseX, double mouseY, int button, CallbackInfoReturnable<Boolean> cir) {
+
+        CreativeModeInventoryScreen self = (CreativeModeInventoryScreen)(Object)this;
+
+        //Dublication support
```
Roelymole: This looks like a typo of duplication.

src/main/java/dev/galacticraft/mod/mixin/client/AbstractContainerScreenAccessor.java
```diff
@@ -0,0 +1,42 @@
...
+    @Accessor("hoveredSlot") void setHoveredSlot(@Nullable Slot slot);
+
+    @Invoker("isHovering")
+    boolean gcIsHovering(Slot slot, double pointX, double pointY);
+
+    @Invoker("getTooltipFromContainerItem")
+    List<Component> gcGetTooltipFromContainerItem(@Nullable ItemStack stack);
+
+    @Invoker("findSlot")
+    Slot gcFindSlot(double x, double y);
```
Roelymole: As far as I can see this invoker is currently unused and this is the only private method with an accessor/invoker here, so you might be able to delete this whole file.

src/main/java/dev/galacticraft/mod/mixin/client/AbstractContainerScreenAccessor.java
```diff
@@ -0,0 +1,42 @@
...
+    @Accessor("imageHeight") int getImageHeight();
+
+    @Invoker("renderSlot")
+    void gcRenderSlot(GuiGraphics graphics, Slot region);
+
+    @Accessor("hoveredSlot") Slot getHoveredSlot();
+    @Accessor("hoveredSlot") void setHoveredSlot(@Nullable Slot slot);
+
+    @Invoker("isHovering")
+    boolean gcIsHovering(Slot slot, double pointX, double pointY);
```
Roelymole: For some reason the method with the signature `isHovering(int, int, int, int, double, double)` is protected, but the method with the signature `isHovering(Slot, double, double)` is private.

The latter is basically just a wrapper for the former, so you could get away with using `isHovering(slot.x, slot.y, 16, 16, mouseX, mouseY)` instead of `isHovering(slot, mouseX, mouseY)`.

src/main/java/dev/galacticraft/mod/mixin/client/CreativeModeInventoryScreenAccessor.java
Roelymole, line 1: I think you should be able to shadow all of these attributes/methods in `CreativeScreenMixin`.

src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
```diff
@@ -0,0 +1,489 @@
...
+                    {
+                        if(slot.getItem().isEmpty())
+                        {
+                            slot.set(carried);
+                            slot.setChanged();
+                            self.getMenu().setCarried(ItemStack.EMPTY);
+                            ClientPlayNetworking.send(new CreativeGcTransferItemPayload(1, slot.getContainerSlot(), 1, carried));
+
+                            Minecraft.getInstance().getSoundManager().play(
+                                    net.minecraft.client.resources.sounds.SimpleSoundInstance.forUI(
```
Roelymole: Please add `import net.minecraft.client.resources.sounds.SimpleSoundInstance;` at the top instead of doing this.

src/main/resources/assets/galacticraft/textures/gui/radiobutton_gear_inventory_buttons.png
Roelymole, line 1: Having such a large texture for this seems quite unnecessary.

src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
```diff
@@ -0,0 +1,489 @@
...
+            {
+                if(gcTryQuickStackToGCInv(region))
+                {
+                    return;
+                }
+            }
+        }
+
+
+    }
```
Roelymole: The following does exactly the same thing much more concisely.

```suggestion
    protected void slotClicked(Slot region, int slotId, int button, ClickType actionType, CallbackInfo ci) {
        if (region != null && actionType == ClickType.QUICK_MOVE) {
            gcTryQuickStackToGCInv(region);
        }
    }
```

src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
```diff
@@ -0,0 +1,489 @@
...
+            else
+            {
+                creativeSwitchButton.setX(-2000);
+                creativeSwitchButton.setY(-2000);
+                creativeSwitchButton.radioButtonOnClick = null;
+            }
+
+        }
+        if(group.getType() != CreativeModeTab.Type.INVENTORY)
+        {
```
Roelymole:
```suggestion
        } else {
```

src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
```diff
@@ -0,0 +1,489 @@
...
+        return bGCInventory;
+    }
+
+    @Override
+    public List<AccessorySlot> gc$getRenderSlots()
+    {
+        return gc$slots;
+    }
+
+    @Unique private List<Slot> GCInvSlots;
```
Roelymole: It is better practice to put all of the attributes at the top of the class before any methods so that they can be found more easily.

src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
```diff
@@ -0,0 +1,489 @@
...
+            );
+            Constructor<?> c = cls.getDeclaredConstructor(Slot.class, int.class, int.class, int.class);
+            c.setAccessible(true);
+            return (Slot) c.newInstance(target, invSlot, x, y);
+        } catch (Exception e) {
+            throw new RuntimeException("Cannot create SlotWrapper", e);
+        }
+    }
+
+    @Unique RadioButton creativeSwitchButton;
```
Roelymole: This should also be at the top with the other attributes.

src/main/java/dev/galacticraft/mod/accessor/GCCreativeGuiSlots.java
```diff
@@ -0,0 +1,15 @@
...
+import net.minecraft.client.gui.GuiGraphics;
+import net.minecraft.world.inventory.Slot;
+
+import java.util.List;
+
+public interface GCCreativeGuiSlots {
+
+    boolean gc$isGCInventoryEnabled();
+    List<AccessorySlot> gc$getRenderSlots();
+    void gc$renderGcSlots(GuiGraphics g, int mouseX, int mouseY);
```
Roelymole, line 28: The single letter name `g` is not ideal, using `guiGraphics` or just `graphics` is much clearer and more readable.

src/main/java/dev/galacticraft/mod/client/gui/widget/RadioButton.java
```diff
@@ -0,0 +1,129 @@
...
+        super(x,y,BTN_WIDTH, BTN_HEIGHT*2, Component.empty());
+    }
+
+    public void setIsBottomButtonActive(boolean newActive)
+    {
+        isBottomButtonActive = newActive;
+    }
+
+    @Override
+    protected void renderWidget(GuiGraphics context, int mouseX, int mouseY, float delta) {
```
Roelymole: I'm not quite sure why you have used the name `context` for an instance of `GuiGraphics`, using `guiGraphics` or just `graphics` would be preferable.

src/main/java/dev/galacticraft/mod/mixin/AbstractContainerMenuMixin.java
Roelymole, line 1: It is probably best to remove this logging before the PR is merged, since we don't really want to use mixins when we don't strictly need them.

src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
```diff
@@ -0,0 +1,489 @@
...
+        int j = topPos + 18;
+        int k = j + 112;
+        if (selectedTab.canScroll()) {
+            ResourceLocation resourceLocation =  ((CreativeModeInventoryScreenAccessor)(Object)this).GCcanScroll() ? ((CreativeModeInventoryScreenAccessor)(Object)this).get_SCROLLER_SPRITE() : ((CreativeModeInventoryScreenAccessor)(Object)this).get_SCROLLER_DISABLED_SPRITE();
+            graphics.blitSprite(resourceLocation, i, j + (int)((float)(k - j - 17) * ((CreativeModeInventoryScreenAccessor)(Object)this).getScrollOffs()), 12, 15);
+        }
+
+        ((CreativeModeInventoryScreenAccessor)(Object)this).GCrenderTabButton(graphics, selectedTab);
+        if (selectedTab.getType() == CreativeModeTab.Type.INVENTORY) {
+            InventoryScreen.renderEntityInInventoryFollowsMouse(graphics, isGCInventoryEnabled() ? leftPos + 54 : leftPos + 73, topPos + 6, leftPos + 105, topPos + 49, 20, 0.0625F, (float)mouseX, (float)mouseY, ((ScreenAccessor)(Object)this).getMinecraft().player);
```
Roelymole: Another little thing that I have just noticed is that the player is not quite centered in the Galacticraft gear inventory.
Shifting the player 1 pixel to the right should be perfect, i.e. `leftPos + 55` instead of `leftPos + 54`.

src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
```diff
@@ -0,0 +1,489 @@
...
+                int l = k / 2;
+                int m = k % 2;
+                n = -2000;
+                j = -2000;
+            } else if (i >= 0 && i < 5) {
+                n = -2000;
+                j = -2000;
+            } else if (i == 45) {
+                n = -2000;
+                j = -2000;
```
Roelymole: These can be combined.

```suggestion
            } else if ((i >= 0 && i < 5) || i == 45) {
                n = -2000;
                j = -2000;
```

src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
```diff
@@ -0,0 +1,489 @@
...
+                playerInv.setItem(i, newStack);
+
+                gcSlot.set(ItemStack.EMPTY);
+                gcSlot.setChanged();
+
+                player.inventoryMenu.broadcastChanges();
+                break;
+            }
+        }
+        return canPlaceInAny;
```
Roelymole: The `canPlaceInAny` variable is superfluous and may be removed by doing the following:

```suggestion
                return true;
            }
        }
        return false;
```

src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
```diff
@@ -0,0 +1,489 @@
...
+                                1.0F
+                        )
+                );
+                playerSlot.set(ItemStack.EMPTY);
+                playerSlot.setChanged();
+                break;
+            }
+        }
+        return canPlaceInAny;
+
```
Roelymole: The `canPlaceInAny` variable is superfluous and may be removed by doing the following:

```suggestion
                return true;
            }
        }
        return false;
```

src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
```diff
@@ -0,0 +1,489 @@
...
+                            slot.setChanged();
+                            self.getMenu().setCarried(ItemStack.EMPTY);
+                            ClientPlayNetworking.send(new CreativeGcTransferItemPayload(1, slot.getContainerSlot(), 1, carried));
+
+                            Minecraft.getInstance().getSoundManager().play(
+                                    net.minecraft.client.resources.sounds.SimpleSoundInstance.forUI(
+                                            SoundEvents.ARMOR_EQUIP_LEATHER,
+                                            1.0F
+                                    )
+                            );
```
Roelymole: It would be much better to call `player.galacticraft$onEquipAccessory(previous, incoming)`.

src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
```diff
@@ -0,0 +1,489 @@
...
+            if( gc$slots.get(i).mayPlace(playerSlot.getItem()) && gc$slots.get(i).getItem().isEmpty())
+            {
+                canPlaceInAny = true;
+                ClientPlayNetworking.send(new CreativeGcTransferItemPayload(1, i, 1, playerSlot.getItem().copy()));
+                Minecraft.getInstance().getSoundManager().play(
+                        net.minecraft.client.resources.sounds.SimpleSoundInstance.forUI(
+                                SoundEvents.ARMOR_EQUIP_LEATHER,
+                                1.0F
+                        )
+                );
```
Roelymole: It would be much better to call `player.galacticraft$onEquipAccessory(previous, incoming)` here too.

src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
```diff
@@ -0,0 +1,489 @@
...
+                creativeSwitchButton.radioButtonOnClick = () ->{
+                    bGCInventory = creativeSwitchButton.getIsBottomButtonActive();
+                    regenerateSlots();
+                };
+                ((ScreenAccessor)(Object)this).gc$addRenderableWidget(creativeSwitchButton);
+            }
+            else
+            {
+                creativeSwitchButton.setX(-2000);
+                creativeSwitchButton.setY(-2000);
```
Roelymole: Setting `creativeSwitchButton.visible = false;` would probably be better than moving it outside of the bounds of the screen.

src/main/java/dev/galacticraft/mod/client/gui/widget/RadioButton.java
```diff
@@ -0,0 +1,126 @@
...
+
+    private boolean isBottomButtonActive = false;
+
+    public boolean getIsBottomButtonActive() {
+        return isBottomButtonActive;
+    }
+
+    public Runnable radioButtonOnClick;
+
+    public RadioButton(int x, int y){
```
Roelymole:
```suggestion
    public RadioButton(int x, int y) {
```

src/main/java/dev/galacticraft/mod/client/gui/widget/RadioButton.java
```diff
@@ -0,0 +1,126 @@
...
+        super(x, y, BTN_WIDTH, BTN_HEIGHT*2, Component.empty());
+    }
+
+    public void setIsBottomButtonActive(boolean newActive)
+    {
```
Roelymole: This is what I was referring to when I said about placing the opening curly brackets on the same line rather than on a new line.

Sorry to be a bother, but please may you go through and make the same fix everywhere else you have done this too.

```suggestion
    public void setIsBottomButtonActive(boolean newActive) {
```

Whilst you are at it please could you quickly add in spaces after commas and if/for/while like I mentioned before.

There are a handful of spaces where they shouldn't be and several extra blank lines too.

It might seem insignificant, but it makes a big difference for readability (and my sanity).

src/main/java/dev/galacticraft/mod/client/gui/widget/RadioButton.java
```diff
@@ -0,0 +1,126 @@
...
+                    this.onClick(mouseX, mouseY);
+                    radioButtonOnClick.run();
+                    return true;
+                }
+
+            }
+            return false;
+        } else {
+            return false;
+        }
```
Roelymole: This can be simplified.

```suggestion
        }
        return false;
```

src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
```diff
@@ -0,0 +1,495 @@
...
+                            slot.setChanged();
+                            getMenu().setCarried(ItemStack.EMPTY);
+                            ClientPlayNetworking.send(new CreativeGcTransferItemPayload(1, slot.getContainerSlot(), 1, carried));
+
+                            Minecraft.getInstance().getSoundManager().play(
+                                    SimpleSoundInstance.forUI(
+                                            SoundEvents.ARMOR_EQUIP_LEATHER,
+                                            1.0F
+                                    )
+                            );
```
Roelymole: It would be much better to call `player.galacticraft$onEquipAccessory(previous, incoming)`.

src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
```diff
@@ -0,0 +1,495 @@
...
+            if( gc$slots.get(i).mayPlace(playerSlot.getItem()) && gc$slots.get(i).getItem().isEmpty())
+            {
+                canPlaceInAny = true;
+                ClientPlayNetworking.send(new CreativeGcTransferItemPayload(1, i, 1, playerSlot.getItem().copy()));
+                Minecraft.getInstance().getSoundManager().play(
+                        SimpleSoundInstance.forUI(
+                                SoundEvents.ARMOR_EQUIP_LEATHER,
+                                1.0F
+                        )
+                );
```
Roelymole: It would be much better to call `player.galacticraft$onEquipAccessory(previous, incoming)` here too.

src/main/java/dev/galacticraft/mod/accessor/GCCreativeGuiSlots.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.util.List;
+
+public interface GCCreativeGuiSlots {
+
+    boolean gc$isGCInventoryEnabled();
+    List<AccessorySlot> gc$getRenderSlots();
+    void gc$renderGcSlots(GuiGraphics graphics, int mouseX, int mouseY);
+
+}
```
Roelymole: I think this is nicer.

```suggestion
public interface GCCreativeGuiSlots {
    boolean gc$isGCInventoryEnabled();

    List<AccessorySlot> gc$getRenderSlots();

    void gc$renderGcSlots(GuiGraphics graphics, int mouseX, int mouseY);
}
```

src/main/java/dev/galacticraft/mod/mixin/client/AbstractContainerScreenMixin.java
```diff
@@ -0,0 +1,52 @@
...
+            at = @At(
+                    value = "INVOKE",
+                    target = "Lnet/minecraft/client/gui/screens/inventory/AbstractContainerScreen;renderLabels(Lnet/minecraft/client/gui/GuiGraphics;II)V",
+                    shift = At.Shift.BEFORE,
+                    ordinal = 0
+            )
+    )
+    private void gc$beforeCarried(GuiGraphics graphics, int mouseX, int mouseY, float delta, CallbackInfo ci) {
+        Object self = this;
+        if (self instanceof GCCreativeGuiSlots gc) {
```
Roelymole: These lines can be combined quite neatly.

```suggestion
        if ((Object) this instanceof GCCreativeGuiSlots gc) {
```

src/main/java/dev/galacticraft/mod/mixin/client/ScreenAccessor.java
Roelymole, line 1: This file can be deleted now as well.

src/main/java/dev/galacticraft/mod/network/GCPackets.java
```diff
@@ -42,5 +42,6 @@ public static void register() {
         PayloadTypeRegistry.playC2S().register(BubbleVisibilityPayload.TYPE, BubbleVisibilityPayload.STREAM_CODEC);
         PayloadTypeRegistry.playC2S().register(PlanetTeleportPayload.TYPE, PlanetTeleportPayload.STREAM_CODEC);
         PayloadTypeRegistry.playC2S().register(OpenGcInventoryPayload.TYPE, OpenGcInventoryPayload.STREAM_CODEC);
+        PayloadTypeRegistry.playC2S().register(CreativeGcTransferItemPayload.TYPE,CreativeGcTransferItemPayload.STREAM_CODEC);
```
Roelymole: A space after the comma here, please.

src/main/java/dev/galacticraft/mod/network/c2s/CreativeGcTransferItemPayload.java
```diff
@@ -0,0 +1,151 @@
...
+import net.minecraft.network.VarInt;
+import org.jetbrains.annotations.NotNull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public record CreativeGcTransferItemPayload(int containerType, int slotIndex, int action, ItemStack stack)
+        implements C2SPayload {
+
+    private static final Logger LOG = LoggerFactory.getLogger(CreativeGcTransferItemPayload.class);
```
Roelymole: This is unused and can be deleted.

src/main/java/dev/galacticraft/mod/network/c2s/CreativeGcTransferItemPayload.java
```diff
@@ -0,0 +1,151 @@
...
+
+    public static final ResourceLocation ID =
+            Constant.id("gc_slot_click");
+
+    public static final Type<CreativeGcTransferItemPayload> TYPE =
+            new Type<>(ID);
```
Roelymole: I think it would be clearer to put these on a single line each, rather than splitting them, since they are not very long lines.

src/main/java/dev/galacticraft/mod/network/c2s/CreativeGcTransferItemPayload.java
```diff
@@ -0,0 +1,151 @@
...
+        for(int i = 0; i < 12; i++)
+        {
+            if(item.is(GCAccessorySlots.SLOT_TAGS.get(i)))
+            {
+                result = true;
+                break;
+            }
+        }
+        return result;
+    }
```
Roelymole: This can be made more concise.

```suggestion
    private boolean canPlaceItem(ItemStack item) {
        for (int i = 0; i < 12; i++) {
            if (item.is(GCAccessorySlots.SLOT_TAGS.get(i))) {
                return true;
            }
        }
        return false;
    }
```

src/main/resources/galacticraft.mixins.json
Roelymole, line 1: The convention here is to put the mixins in alphabetical order, so `"client.AbstractContainerScreenMixin"` would go after `"client.AbstractClientPlayerMixin"` and `"client.CreativeScreenMixin"` would go after `"client.ClientPacketListenerMixin"`.

`"client.ScreenAccessor"` isn't needed anymore and should be deleted.

src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
```diff
@@ -0,0 +1,484 @@
...
+                if (slot != null)
+                {
+                    //Quick stack
+                    if (Screen.hasShiftDown())
+                    {
+                        gcTryQuickStackToPlayerInv(slot);
+                        cir.setReturnValue(true);
+                        return;
+                    }
+                    if (!carried.isEmpty() && slot.mayPlace(carried))
```
Roelymole: You could swap the order of this if statement and the corresponding else if statement around to simplify things slightly.

```java
if (carried.isEmpty()) {
    ...
} else if (slot.mayPlace(carried)) {
    // We know that !carried.isEmpty() is true without needing to check
    ...
}
```

src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
```diff
@@ -0,0 +1,484 @@
...
+        }
+        return canPlaceInAny;
+
+    }
+
+    @Inject(method = "slotClicked", at = @At("HEAD"))
+    protected void slotClicked(Slot region, int slotId, int button, ClickType actionType, CallbackInfo ci)
+    {
+        if (region != null && actionType == ClickType.QUICK_MOVE) {
+            if (gcTryQuickStackToGCInv(region)) { return; };
```
Roelymole: All you need is the part in the condition for the if statement.

```suggestion
            gcTryQuickStackToGCInv(region);
```

src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
```diff
@@ -0,0 +1,484 @@
...
+            int row = i/4;
+            row = 1 - row;
+            generateGCSlot(99+col*18, (row*27)+33, i);
+        }
+    }
+
+    @Unique
+    private void generateGCSlot(int x, int y, int idx)
+    {
+        AccessorySlot s = new AccessorySlot(
```
Roelymole: There isn't much point having the temporary variable `AccessorySlot s` if it is only used once.

```suggestion
        gc$slots.add(new AccessorySlot(
```

src/main/java/dev/galacticraft/mod/client/gui/widget/RadioButton.java
Roelymole: I guess this was unused/no longer necessary?

src/main/generated/assets/galacticraft/lang/en_us.json
Roelymole: It looks like you haven't actually run the datagen (the space before the colon gives you away).

I would encourage you to do this to get a feel for how the runDatagen Gradle task works as well as to check that everything is correct.

src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
```diff
@@ -0,0 +1,484 @@
...
+    }
+
+    @Inject(method = "init", at = @At("HEAD"))
+    private void init(CallbackInfo ci)
+    {
+        creativeSwitchButton = new RadioButton(0,0);
+    }
+
+    @Inject(method = "selectTab", at = @At(value = "HEAD"))
+    void gc$selectTab(CreativeModeTab group, CallbackInfo ci)
```
Roelymole: It would be clearer and more consistent if you marked this method and the next method as private, thanks!

src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
```diff
@@ -0,0 +1,484 @@
...
+                addRenderableWidget(creativeSwitchButton);
+            } else {
+                creativeSwitchButton.setX(-2000);
+                creativeSwitchButton.setY(-2000);
+                creativeSwitchButton.radioButtonOnClick = null;
+            }
+
+        }
+        if (group.getType() != CreativeModeTab.Type.INVENTORY)
+        {
```
Roelymole: I feel like I mentioned this one before... maybe it got lost.

```suggestion
        } else {
```

src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
```diff
@@ -0,0 +1,369 @@
...
+
+        for (int i = 0; i < 8; i++) {
+            int column = i % 4;
+            int row = 1 - (i / 4);
+            generateGCSlot(99 + column * 18, yOffset + row * 27, i);
+        }
+
+        for (int i = 0; i < 4; i++) {
+            int column = i % 2;
+            int row = i / 2;
```
Roelymole: These should be the other way round to match the positioning of the vanilla armour slots. The problem before was that the names were the wrong way round, the values were correct.

```suggestion
            int column = i / 2;
            int row = i % 2;
```

src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
```diff
@@ -0,0 +1,369 @@
...
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+
+import java.lang.reflect.Constructor;
+import java.util.ArrayList;
+import java.util.List;
+
+@Mixin(CreativeModeInventoryScreen.class)
+public abstract class CreativeScreenMixin extends EffectRenderingInventoryScreen implements GCCreativeGuiSlots {
+
+    private static final ResourceLocation GC_GUIBG_TEX = Constant.id("textures/gui/creative_tab_inventory.png");
```
Roelymole: This should be marked as `@Unique` too.

src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
Roelymole: Of course, `AccessorySlot` was already calling this anyway.

src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
```diff
@@ -0,0 +1,369 @@
...
+                    bGCInventory = creativeSwitchButton.getIsBottomButtonActive();
+                    regenerateSlots();
+                };
+                addRenderableWidget(creativeSwitchButton);
+            } else {
+                creativeSwitchButton.visible = false;
+                creativeSwitchButton.radioButtonOnClick = null;
+            }
+        }
+        if (group.getType() != CreativeModeTab.Type.INVENTORY) {
```
Roelymole:
```suggestion
        } else {
```

src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
```diff
@@ -0,0 +1,369 @@
...
+                        ItemStack slotStack = slot.getItem();
+                        getMenu().setCarried(slotStack);
+                        slot.set(ItemStack.EMPTY);
+                        slot.setChanged();
+                        ClientPlayNetworking.send(new CreativeGcTransferItemPayload(1, slot.getContainerSlot(), 0, ItemStack.EMPTY));
+                        cir.setReturnValue(true);
+                        return;
+                    }
+                    cir.setReturnValue(true);
+                    return;
```
Roelymole, line 182: This should do the same thing without so much repeated code.

```suggestion
                    if (Screen.hasShiftDown()) {
                        gcTryQuickStackToPlayerInv(slot);
                    } else if (carried.isEmpty() || slot.mayPlace(carried)) {
                        getMenu().setCarried(slot.getItem());
                        slot.set(carried);
                        slot.setChanged();
                        ClientPlayNetworking.send(new CreativeGcTransferItemPayload(1, slot.getContainerSlot(), carried.isEmpty() ? 0 : 1, carried));
                    }
                    cir.setReturnValue(true);
                    return;
```

src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
```diff
@@ -0,0 +1,369 @@
...
+        if (selectedTab.canScroll()) {
+            ResourceLocation resourceLocation = canScroll() ? SCROLLER_SPRITE : SCROLLER_DISABLED_SPRITE;
+            graphics.blitSprite(resourceLocation, i, j + (int)((float)(k - j - 17) * scrollOffs), 12, 15);
+        }
+
+        renderTabButton(graphics, selectedTab);
+        if (selectedTab.getType() == CreativeModeTab.Type.INVENTORY) {
+            InventoryScreen.renderEntityInInventoryFollowsMouse(graphics, isGCInventoryEnabled() ? leftPos + 55 : leftPos + 73, topPos + 6, leftPos + 105, topPos + 49, 20, 0.0625F, (float)mouseX, (float)mouseY, minecraft.player);
+        }
+    }
```
Roelymole: `@Overwrite` should only be used as an absolute last resort. Modifying two arguments is all that is needed: the first to change the background texture and the second to shift the entity to the left.

You'll need to add `import org.spongepowered.asm.mixin.injection.ModifyArg;` at the top.

```suggestion
    @ModifyArg(method = "renderBg", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/gui/GuiGraphics;blit(Lnet/minecraft/resources/ResourceLocation;IIIIII)V"), index = 0)
    private ResourceLocation gc$changeBackground(ResourceLocation original) {
        return isGCInventoryEnabled() ? GC_GUIBG_TEX : original;
    }

    @ModifyArg(method = "renderBg", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/gui/screens/inventory/InventoryScreen;renderEntityInInventoryFollowsMouse(Lnet/minecraft/client/gui/GuiGraphics;IIIIIFFFLnet/minecraft/world/entity/LivingEntity;)V"), index = 1)
    private int gc$shiftEntityLeft(int original) {
        return isGCInventoryEnabled() ? leftPos + 55 : original;
    }
```

src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
```diff
@@ -0,0 +1,324 @@
...
+
+    @Shadow protected abstract void renderTabButton(GuiGraphics graphics, CreativeModeTab group);
+
+    @Shadow private EditBox searchBox;
+
+    @Shadow protected abstract boolean canScroll();
+
+    @Shadow @Final private static ResourceLocation SCROLLER_SPRITE;
+    @Shadow @Final private static ResourceLocation SCROLLER_DISABLED_SPRITE;
+    @Shadow private float scrollOffs;
```
Roelymole: We don't need these anymore. I expect that there will be a few unused imports after removing them, so make sure to rerun spotlessApply.

src/main/java/dev/galacticraft/mod/mixin/client/CreativeScreenMixin.java
```diff
@@ -0,0 +1,324 @@
...
+        } catch (Exception e) {
+            throw new RuntimeException("Cannot create SlotWrapper", e);
+        }
+    }
+
+    @ModifyArg(method = "renderBg", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/gui/GuiGraphics;blit(Lnet/minecraft/resources/ResourceLocation;IIIIII)V"), index = 0)
+    private ResourceLocation gc$changeBackground(ResourceLocation original) {
+        return isGCInventoryEnabled() ? GC_GUIBG_TEX : original;
+    }
+    @ModifyArg(method = "renderBg", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/gui/screens/inventory/InventoryScreen;renderEntityInInventoryFollowsMouse(Lnet/minecraft/client/gui/GuiGraphics;IIIIIFFFLnet/minecraft/world/entity/LivingEntity;)V"), index = 1)
```
Roelymole, line 112: Please could you put a new line between these two methods just to make them more distinct.

src/main/java/dev/galacticraft/mod/client/gui/widget/RadioButton.java
```diff
@@ -0,0 +1,111 @@
...
+
+    public int getButtonHovered(int mouseX, int mouseY) {
+        boolean top = mouseX >= getX() && mouseX < getX() + BTN_WIDTH &&
+                mouseY >= getY() && mouseY < getY() + BTN_HEIGHT;
+
+        boolean bottom = mouseX >= getX() && mouseX < getX() + BTN_WIDTH &&
+                mouseY >= getY() + BTN_HEIGHT && mouseY < getY() + BTN_HEIGHT * 2;
+        if (top) {
+            return 0;
+        }
```
Roelymole: It would be slightly more efficient to swap the order of these around to avoid having to calculate bottom if top is true and we are going to return 0 anyway.

Roelymole (approved):
Thank you very much for all the hard work!

PR #544, opened 2026-01-31
Rocket upgrades support explosive blocks now.

maxryan008:
Rockets can have an explosive payload that explodes on collision.

### DIFF
--- a/src/main/java/dev/galacticraft/api/rocket/RocketData.java
+++ b/src/main/java/dev/galacticraft/api/rocket/RocketData.java
@@ -214,8 +214,8 @@ public record RocketData(
         return Optional.empty();
     }
 
-    public boolean hasUpgradeData() {
-        return this.upgradeData.isPresent();
+    public <T extends RocketUpgradeData> boolean hasUpgradeData(Class<T> clazz) {
+        return this.upgradeData.isPresent() && clazz.isInstance(this.upgradeData.get());
     }
 
     private static <T> Optional<EitherHolder<T>> maybeHolder(@Nullable Holder<T> holder) {
@@ -236,5 +236,9 @@ public record RocketData(
         } else {
             builder.remove(type);
         }
+    }
+
+    public boolean hasUpgrade(ResourceKey<RocketUpgrade<?,?>> upgrade) {
+        return this.upgrade.isPresent() && this.upgrade.get().key().equals(upgrade);
     }
 }
--- a/src/main/java/dev/galacticraft/mod/content/entity/vehicle/RocketEntity.java
+++ b/src/main/java/dev/galacticraft/mod/content/entity/vehicle/RocketEntity.java
@@ -37,6 +37,7 @@ import dev.galacticraft.mod.api.block.entity.FuelDock;
 import dev.galacticraft.mod.attachments.GCServerPlayer;
 import dev.galacticraft.mod.content.GCBlocks;
 import dev.galacticraft.mod.content.GCFluids;
+import dev.galacticraft.mod.content.GCRocketParts;
 import dev.galacticraft.mod.content.GCStats;
 import dev.galacticraft.mod.content.advancements.GCTriggers;
 import dev.galacticraft.mod.content.block.special.launchpad.AbstractLaunchPad;
@@ -95,7 +96,6 @@ import net.minecraft.world.phys.Vec3;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 
@@ -346,7 +346,9 @@ public class RocketEntity extends AdvancedVehicle implements Rocket, IgnoreShift
 
     @Override
     public void move(MoverType type, Vec3 vec3d) {
-        if (onGround()) vec3d = vec3d.multiply(1.0D, 0.0D, 1.0D);
+        if (onGround()) {
+            vec3d = vec3d.multiply(1.0D, 0.0D, 1.0D);
+        }
         super.move(type, vec3d);
         this.getPassengers().forEach(this::positionRider);
     }
@@ -418,8 +420,8 @@ public class RocketEntity extends AdvancedVehicle implements Rocket, IgnoreShift
 
     private boolean hasExplosiveUpgradeAndPayload() {
         RocketData data = this.getRocketData();
-        if (data.getUpgradeData(ExplosiveRocketData.class).isEmpty()) return false;
-        if (data.upgrade().isEmpty()) return false;
+        if (!data.hasUpgradeData(ExplosiveRocketData.class)) return false;
+        if (!data.hasUpgrade(GCRocketParts.EXPLOSIVE_UPGRADE)) return false;
 
         ResourceKey<?> key = data.upgrade().get().key();
 
@@ -508,7 +510,7 @@ public class RocketEntity extends AdvancedVehicle implements Rocket, IgnoreShift
 
         BlockState oldState = level.getBlockState(pos);
         BlockState payloadState = tntBlock.defaultBlockState();
-        level.setBlock(pos, payloadState, 3);
+        level.setBlock(pos, payloadState, Block.UPDATE_ALL);
 
         try {
             Explosion dummy = level.explode(
@@ -528,19 +530,17 @@ public class RocketEntity extends AdvancedVehicle implements Rocket, IgnoreShift
                     this.getBoundingBox().inflate(scanRadius)
             );
 
-            List<PrimedTnt> spawned = new ArrayList<>();
-            for (PrimedTnt e : after) {
-                if (!before.contains(e)) spawned.add(e);
+            boolean spawned = false;
+            for (PrimedTnt primed : after) {
+                if (!before.contains(primed)) {
+                    spawned = true;
+                    primed.setFuse(0);
+                    primed.setDeltaMovement(Vec3.ZERO);
+                    primed.tick();
+                }
             }
 
-            for (PrimedTnt primed : spawned) {
-                primed.setFuse(0);
-                primed.setDeltaMovement(Vec3.ZERO);
-
-                primed.tick();
-            }
-
-            if (spawned.isEmpty()) {
+            if (!spawned) {
                 LivingEntity igniter = (this.getFirstPassenger() instanceof LivingEntity le) ? le : null;
                 PrimedTnt primed = new PrimedTnt(level, this.getX(), this.getY(), this.getZ(), igniter);
                 primed.setFuse(0);
@@ -548,7 +548,7 @@ public class RocketEntity extends AdvancedVehicle implements Rocket, IgnoreShift
                 primed.tick();
             }
         } finally {
-            level.setBlock(pos, oldState, 3);
+            level.setBlock(pos, oldState, Block.UPDATE_ALL);
         }
     }
 
@@ -562,7 +562,7 @@ public class RocketEntity extends AdvancedVehicle implements Rocket, IgnoreShift
         boolean collided = this.horizontalCollision || this.verticalCollision || this.isInWall();
 
         double speedSqr = preMoveDelta.lengthSqr();
-        double horizSqr = (preMoveDelta.x * preMoveDelta.x) + (preMoveDelta.z * preMoveDelta.z);
+        double horizSqr = preMoveDelta.horizontalDistanceSqr();
         double upSqr = (preMoveDelta.y > 0) ? (preMoveDelta.y * preMoveDelta.y) : 0.0;
 
         if (collided && speedSqr >= CRASH_MIN_SPEED_SQR) {
--- a/src/main/java/dev/galacticraft/mod/screen/RocketWorkbenchMenu.java
+++ b/src/main/java/dev/galacticraft/mod/screen/RocketWorkbenchMenu.java
@@ -35,6 +35,7 @@ import dev.galacticraft.mod.content.rocket.part.data.ExplosiveRocketData;
 import dev.galacticraft.mod.content.rocket.part.data.RocketUpgradeData;
 import dev.galacticraft.mod.machine.storage.VariableSizedContainer;
 import dev.galacticraft.mod.recipe.RocketRecipe;
+import dev.galacticraft.mod.tag.GCItemTags;
 import dev.galacticraft.mod.world.inventory.RocketResultSlot;
 import io.netty.buffer.ByteBuf;
 import net.minecraft.core.BlockPos;
@@ -219,7 +220,7 @@ public class RocketWorkbenchMenu extends AbstractContainerMenu implements Variab
         if (upgradeStack.getItem() instanceof BlockItem blockItem && blockItem.getBlock() instanceof TntBlock tntBlock) {
             upgrade = Optional.of(new EitherHolder<>(GCRocketParts.EXPLOSIVE_UPGRADE));
             upgradeData = Optional.of(new ExplosiveRocketData(BuiltInRegistries.BLOCK.getKey(tntBlock)));
-        } else if (!upgradeStack.isEmpty()) {
+        } else if (upgradeStack.is(GCItemTags.ROCKET_STORAGE_UPGRADE_ITEMS)) {
             upgrade = Optional.of(new EitherHolder<>(GCRocketParts.STORAGE_UPGRADE));
         }
 


src/main/java/dev/galacticraft/api/component/GCDataComponents.java
```diff
@@ -39,8 +39,14 @@
...
-    public static final DataComponentType<RocketData> ROCKET_DATA = register("rocket_data", b -> b
-            .persistent(RocketData.CODEC).networkSynchronized(RocketData.STREAM_CODEC));
+    public static final DataComponentType<RocketData> ROCKET_DATA = Registry.register(
+            BuiltInRegistries.DATA_COMPONENT_TYPE,
+            Constant.id("rocket_data"),
+            DataComponentType.<RocketData>builder()
+                    .persistent(RocketData.CODEC)
+                    .networkSynchronized(RocketData.STREAM_CODEC)
+                    .build()
+    );
```
Roelymole, line 49: Does this fix something? I can't see anything particularly different between the old and the new version.

src/main/java/dev/galacticraft/api/rocket/part/RocketUpgrade.java
```diff
@@ -45,8 +46,8 @@ public non-sealed interface RocketUpgrade<C extends RocketUpgradeConfig, T exten
     Codec<HolderSet<RocketUpgrade<?, ?>>> LIST_CODEC = RegistryCodecs.homogeneousList(RocketRegistries.ROCKET_UPGRADE, DIRECT_CODEC);
     StreamCodec<RegistryFriendlyByteBuf, Holder<RocketUpgrade<?, ?>>> STREAM_CODEC = StreamCodecs.ofHolder(RocketRegistries.ROCKET_UPGRADE);
 
-    Codec<EitherHolder<RocketUpgrade<?, ?>>> EITHER_CODEC = EitherHolder.codec(RocketRegistries.ROCKET_UPGRADE, CODEC);
-    StreamCodec<RegistryFriendlyByteBuf, EitherHolder<RocketUpgrade<?, ?>>> EITHER_STREAM_CODEC = EitherHolder.streamCodec(RocketRegistries.ROCKET_UPGRADE, STREAM_CODEC);
+    public static final Codec<EitherHolder<RocketUpgrade<?, ?>>> EITHER_CODEC =
```
Roelymole: The `public static final` declaration is redundant in an interface and these keywords are omitted in most/all interfaces in the codebase.

src/main/java/dev/galacticraft/mod/content/block/entity/RocketWorkbenchBlockEntity.java
```diff
@@ -55,6 +58,12 @@ public boolean canPlaceItem(int index, ItemStack stack) {
     public final SimpleContainer chests = new SimpleContainer(1) {
         @Override
         public boolean canPlaceItem(int slot, ItemStack stack) {
+            if (!(stack.getItem() instanceof BlockItem bi)) return false;
+
+            Block b = bi.getBlock();
+
+            if (b instanceof TntBlock) return true;
```
Roelymole: Returning false straight away would prevent any non-block items from being used, even if they are part of the `GCItemTags.ROCKET_STORAGE_UPGRADE_ITEMS` tag (which isn't a problem with the default items in this tag, but mod packs might add to this).

```suggestion
            if (stack.getItem() instanceof BlockItem blockItem && blockItem.getBlock() instanceof TntBlock) {
                return true;
            }
```

src/main/java/dev/galacticraft/api/rocket/part/RocketUpgradeData.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    public static @NotNull RocketUpgradeData forExplosiveBlock(@NotNull Block block) {
+        ResourceLocation id = BuiltInRegistries.BLOCK.getKey(block);
+        return new RocketUpgradeData(Optional.ofNullable(id));
+    }
+
+    public static @Nullable RocketUpgradeData fromStackIfBlock(@NotNull ItemStack stack) {
+        if (!(stack.getItem() instanceof BlockItem bi)) return null;
+        return forExplosiveBlock(bi.getBlock());
```
Roelymole: I think something like this is more readable.

```suggestion
        if (stack.getItem() instanceof BlockItem blockItem) {
            return forExplosiveBlock(blockItem.getBlock());
        }
        return null;
```

src/main/java/dev/galacticraft/api/rocket/RocketCodecs.java
```diff
@@ -0,0 +1,38 @@
...
+import net.minecraft.resources.ResourceKey;
+import net.minecraft.world.item.EitherHolder;
+
+public final class RocketCodecs {
+    private RocketCodecs() {}
+
+    public static <T> Codec<EitherHolder<T>> eitherHolderWithRegistry(ResourceKey<? extends Registry<T>> registryKey) {
+        Codec<ResourceKey<T>> keyCodec = ResourceKey.codec((ResourceKey<Registry<T>>) registryKey);
+
+        return keyCodec.xmap(EitherHolder::new, EitherHolder::key);
```
Roelymole: I think this would suffice.

```suggestion
        return ResourceKey.codec(registryKey).xmap(EitherHolder::new, EitherHolder::key);
```

Also, there is currently only one usage of this, so it might be worth considering whether it is likely to be used elsewhere. If it is not, then perhaps this code could be moved to the `RocketUpgrade` class where it is used. Otherwise, there is nothing necessarily specific to rockets here, so it might be better to put this somewhere more generic.

src/main/java/dev/galacticraft/mod/screen/RocketWorkbenchMenu.java
```diff
@@ -208,17 +207,64 @@ public int getRecipeSize() {
         return out;
     }
 
+    private boolean isExplosiveUpgradeItem(ItemStack stack) {
+        if (!(stack.getItem() instanceof BlockItem bi)) return false;
+        Block block = bi.getBlock();
+        return block instanceof TntBlock;
```
Roelymole: This could be a one-liner. I think it reads clearer too.

```suggestion
        return stack.getItem() instanceof BlockItem blockItem && blockItem.getBlock() instanceof TntBlock;
```

src/main/java/dev/galacticraft/mod/screen/RocketWorkbenchMenu.java
```diff
@@ -208,17 +207,64 @@ public int getRecipeSize() {
...
+        }
+        return Optional.of(new EitherHolder<>(GCRocketParts.STORAGE_UPGRADE));
+    }
+
+    private Optional<ResourceLocation> computeExplosiveBlockId(ItemStack upgradeStack) {
+        if (upgradeStack.isEmpty()) return Optional.empty();
+        if (!(upgradeStack.getItem() instanceof BlockItem bi)) return Optional.empty();
+        if (!isExplosiveUpgradeItem(upgradeStack)) return Optional.empty();
+
+        return Optional.of(net.minecraft.core.registries.BuiltInRegistries.BLOCK.getKey(bi.getBlock()));
```
Roelymole: Import `BuiltInRegistries` at the top, please.

src/main/java/dev/galacticraft/mod/screen/RocketWorkbenchMenu.java
```diff
@@ -208,17 +207,64 @@ public int getRecipeSize() {
...
+
+    private RocketData withWorkbenchUpgrade(RocketData base) {
+        ItemStack upgradeStack = this.workbench.chests.getItem(0);
+
+        Optional<EitherHolder<RocketUpgrade<?, ?>>> key = computeUpgradeKey(upgradeStack);
+        Optional<ResourceLocation> explosive = computeExplosiveBlockId(upgradeStack);
+
+        if (key.isEmpty() || !isExplosiveUpgradeItem(upgradeStack)) {
+            explosive = Optional.empty();
+        }
```
Roelymole: The `isExplosiveUpgradeItem`, `computeUpgradeKey` and `computeExplosiveBlockId` methods are quite redundant and can all be deleted and replaced by something like the following:

```suggestion
        Optional<EitherHolder<RocketUpgrade<?, ?>>> key = Optional.empty();
        Optional<ResourceLocation> explosive = Optional.empty();

        if (stack.getItem() instanceof BlockItem blockItem && blockItem.getBlock() instanceof TntBlock tntBlock) {
            key = Optional.of(new EitherHolder<>(GCRocketParts.EXPLOSIVE_UPGRADE));
            explosive = Optional.of(BuiltInRegistries.BLOCK.getKey(tntBlock));
        } else if (!upgradeStack.isEmpty()) {
            key = Optional.of(new EitherHolder<>(GCRocketParts.STORAGE_UPGRADE));
        }
```

src/main/java/dev/galacticraft/api/rocket/RocketData.java
```diff
@@ -51,6 +52,7 @@ public record RocketData(
         Optional<EitherHolder<RocketBooster<?, ?>>> booster,
         Optional<EitherHolder<RocketEngine<?, ?>>> engine,
         Optional<EitherHolder<RocketUpgrade<?, ?>>> upgrade,
+        Optional<ResourceLocation> explosiveBlock,
```
Roelymole: Can this not be included as part of `upgrade`?

src/main/java/dev/galacticraft/mod/content/item/RocketItem.java
```diff
@@ -114,8 +114,29 @@ public void appendHoverText(ItemStack stack, TooltipContext context, List<Compon
...
+            list.add(
+                    RocketPartTypes.UPGRADE.name.copy()
+                            .append(": ")
+                            .append(upgradeName)
+                            .withStyle(style)
+            );
+
+            if (upgradeKey.equals(dev.galacticraft.mod.content.GCRocketParts.EXPLOSIVE_UPGRADE)) {
+                data.explosiveBlock().ifPresent(id -> {
+                    var block = net.minecraft.core.registries.BuiltInRegistries.BLOCK.get(id);
```
Roelymole: Import `GCRocketParts` and `BuiltInRegistries` at the top, please.

src/main/java/dev/galacticraft/api/rocket/RocketData.java
```diff
@@ -158,11 +208,8 @@ private static <T> Optional<EitherHolder<T>> maybeHolder(@Nullable ResourceKey<T
...
 
-    private <T> void maybeSet(DataComponentPatch.Builder builder, DataComponentType<T> type, @Nullable T value) {
-        if (value != null) {
-            builder.set(type, value);
-        } else {
-            builder.remove(type);
-        }
+    private static <T> void maybeSet(DataComponentPatch.Builder builder, DataComponentType<T> type, @Nullable T value) {
+        if (value != null) builder.set(type, value);
+        else builder.remove(type);
```
Roelymole: I'm really not keen on removing the curly brackets from this if else statement.

src/main/java/dev/galacticraft/api/rocket/part/RocketUpgradeData.java
```diff
@@ -0,0 +1,57 @@
...
+    public boolean hasExplosive() {
+        return this.explosiveBlock.isPresent();
+    }
+
+    public static @NotNull RocketUpgradeData forExplosiveBlock(@NotNull Block block) {
+        ResourceLocation id = BuiltInRegistries.BLOCK.getKey(block);
+        return new RocketUpgradeData(Optional.ofNullable(id));
+    }
+
+    public static @Nullable RocketUpgradeData fromStackIfBlock(@NotNull ItemStack stack) {
```
Roelymole: This method appears to be unused and the only place that the `forExplosiveBlock` method is used is here too.

src/main/java/dev/galacticraft/api/rocket/RocketData.java
```diff
@@ -71,51 +84,89 @@ public RocketData(
...
+        );
     }
 
-    public int alpha() {
-        return FastColor.ARGB32.alpha(this.color());
-    }
+    public int red()   { return FastColor.ARGB32.red(this.color()); }
+    public int green() { return FastColor.ARGB32.green(this.color()); }
+    public int blue()  { return FastColor.ARGB32.blue(this.color()); }
+    public int alpha() { return FastColor.ARGB32.alpha(this.color()); }
```
Roelymole: I personally prefered the way these were before, rather than squashing them onto a single line each.

src/main/java/dev/galacticraft/mod/screen/RocketWorkbenchMenu.java
```diff
@@ -131,20 +133,17 @@ protected void addSlots() {
...
         this.addSlot(new RocketResultSlot(this, this.workbench.output, 0, OUTPUT_X, OUTPUT_Y));
 
-        // Player inventory
         for (int row = 0; row < 3; ++row) {
             for (int column = 0; column < 9; ++column) {
                 this.addSlot(new Slot(this.playerInventory, column + row * 9 + 9, column * 18 + 8, row * 18 + 167));
             }
         }
 
-        // Player hotbar
```
Roelymole, line 147: I don't think it is immediately obvious that these are the hotbar slots, so this comment was quite helpful.

src/main/java/dev/galacticraft/mod/content/entity/vehicle/RocketEntity.java
```diff
@@ -558,9 +740,9 @@ public void tickInAir() {
         double velX = horizontal * sinPitch * Mth.sin(this.getYRot() * Mth.DEG_TO_RAD);
         double velZ = horizontal * -sinPitch * Mth.cos(this.getYRot() * Mth.DEG_TO_RAD);
 
-        // The coefficient of this.getDeltaMovement().y() controls the terminal velocity
-        // You might have to solve a differential equation to obtain a specific value
```
Roelymole, line 562: This comment is also clarifying something that is not necessarily obvious, so I would rather it is kept in.

Roelymole (changes requested):
Thanks for fixing the issues I mentioned!

I haven't had a chance to go over all of the changes in the RocketEntity class properly, partly because the diff isn't the clearest, so I will try to get back to that at some point.

src/main/java/dev/galacticraft/api/component/GCDataComponents.java
maxryan008: No functionality change, it just made it slightly nicer to look at in my opinion?

src/main/java/dev/galacticraft/api/rocket/part/RocketUpgradeData.java
maxryan008: removed

src/main/java/dev/galacticraft/api/rocket/RocketData.java
106566540: So it cant be included as part of upgrade but I have made it a little more configurable by designing a RocketUpgradeData system that allows custom RocketUpgradeData. Currently its only purpose is for the explosive blocks but in the future if we have other upgrades that are semi unique similar to how the explosive upgrade can take any explosive block, then the upgrade data class can be adapted to also allow other data to be stored. for example its not so useful for things such as engine types because engine types there might only be 3 or 4 of so usually it would be stored as a registered upgrade type similar to how chests are done, but perhaps something like a decal such as flags that have a unique identifier and would be impractical to register hundreds of flag types would work well because the rocket data can take variables that are allowed to change. The reason it cant be included in upgrade is due to how RocketUpgrade works. It is a registered upgrade that gets statically registered in GCRocketParts so once its been registered it cant be changed. If you want to make it included in the RocketUpgrade class then we would need to register every single explosive block at game start as their own unique ExplosiveRocketUpgrade instances. Anyway I think the compromise I have designed should suit well because its expandable now and not just a simple ResourceLocation.

maxryan008: Whoops that comment was me. Used the wrong account smh

src/main/java/dev/galacticraft/mod/content/entity/vehicle/RocketEntity.java
```diff
@@ -332,7 +346,7 @@ public void setInitialYRot(float yaw) {
 
     @Override
     public void move(MoverType type, Vec3 vec3d) {
-        if (onGround()) vec3d.multiply(1.0D, 0.0D, 1.0D);
+        if (onGround()) vec3d = vec3d.multiply(1.0D, 0.0D, 1.0D);
```
Roelymole: It might be clearer to put this assignment on the next line inside curly brackets.

src/main/java/dev/galacticraft/api/rocket/RocketData.java
```diff
@@ -142,10 +204,20 @@ public boolean canTravel(@NotNull HolderLookup.Provider lookup, CelestialBody<?,
...
+    public <T extends RocketUpgradeData> Optional<T> getUpgradeData(Class<T> clazz) {
+        if (this.upgradeData.isPresent() && clazz.isInstance(this.upgradeData.get())) {
+            return Optional.of(clazz.cast(this.upgradeData.get()));
+        }
+        return Optional.empty();
+    }
+
+    public boolean hasUpgradeData() {
+        return this.upgradeData.isPresent();
+    }
```
Roelymole: This method is currently unused, however in the `hasExplosiveUpgradeAndPayload` method of `RocketEntity` you check if `data.getUpgradeData(ExplosiveRocketData.class).isEmpty()` but don't actually care about what the value of the upgrade data is, so it might be worth adding something like this:

```suggestion
    public <T extends RocketUpgradeData> boolean hasUpgradeData(Class<T> clazz) {
        return this.upgradeData.isPresent() && clazz.isInstance(this.upgradeData.get());
    }
```

Then `getUpgradeData` can also use this method in the condition for the if statement, which should make it a bit more readable.

src/main/java/dev/galacticraft/mod/content/entity/vehicle/RocketEntity.java
```diff
@@ -393,6 +407,176 @@ public void removePassenger(Entity entity) {
...
+        this.crashArmTimer = 0;
+    }
+
+    private boolean isCrashArmed() {
+        return this.crashArmTimer >= CRASH_ARM_DELAY_TICKS;
+    }
+
+    private boolean hasExplosiveUpgradeAndPayload() {
+        RocketData data = this.getRocketData();
+        if (data.getUpgradeData(ExplosiveRocketData.class).isEmpty()) return false;
```
Roelymole: This is where `getUpgradeData` could be replaced by the version of `hasUpgradeData` I suggested.

```suggestion
        if (!data.hasUpgradeData(ExplosiveRocketData.class)) return false;
```

src/main/java/dev/galacticraft/mod/content/entity/vehicle/RocketEntity.java
```diff
@@ -393,6 +407,176 @@ public void removePassenger(Entity entity) {
...
+        if (this.crashed) return;
+        if (this.getLaunchStage().ordinal() < LaunchStage.LAUNCHED.ordinal()) return;
+
+        this.crashArmTimer++;
+        if (!isCrashArmed()) return;
+
+        boolean collided = this.horizontalCollision || this.verticalCollision || this.isInWall();
+
+        double speedSqr = preMoveDelta.lengthSqr();
+        double horizSqr = (preMoveDelta.x * preMoveDelta.x) + (preMoveDelta.z * preMoveDelta.z);
```
Roelymole:
```suggestion
        double horizSqr = preMoveDelta.horizontalDistanceSqr();
```

src/main/java/dev/galacticraft/mod/screen/RocketWorkbenchMenu.java
```diff
@@ -208,17 +210,44 @@ public int getRecipeSize() {
...
+    private RocketData withWorkbenchUpgrade(RocketData base) {
+        ItemStack upgradeStack = this.workbench.chests.getItem(0);
+
+        Optional<EitherHolder<RocketUpgrade<?, ?>>> upgrade = Optional.empty();
+        Optional<RocketUpgradeData> upgradeData = Optional.empty();
+
+        if (upgradeStack.getItem() instanceof BlockItem blockItem && blockItem.getBlock() instanceof TntBlock tntBlock) {
+            upgrade = Optional.of(new EitherHolder<>(GCRocketParts.EXPLOSIVE_UPGRADE));
+            upgradeData = Optional.of(new ExplosiveRocketData(BuiltInRegistries.BLOCK.getKey(tntBlock)));
+        } else if (!upgradeStack.isEmpty()) {
```
Roelymole: We should probably still check that the item stack is included in the `#galacticraft:rocket_storage_upgrade_items` item tag.

```suggestion
        } else if (upgradeStack.is(GCItemTags.ROCKET_STORAGE_UPGRADE_ITEMS)) {
```

src/main/java/dev/galacticraft/mod/content/entity/vehicle/RocketEntity.java
```diff
@@ -393,6 +407,176 @@ public void removePassenger(Entity entity) {
...
+                    false,
+                    Level.ExplosionInteraction.NONE
+            );
+
+            tntBlock.wasExploded(level, pos, dummy);
+
+            List<PrimedTnt> after = level.getEntitiesOfClass(
+                    PrimedTnt.class,
+                    this.getBoundingBox().inflate(scanRadius)
+            );
```
Roelymole, line 531: Is this setting off an explosion which doesn't do any damage, but ignites other nearby TNT blocks?

Then the next part skips countdown, making them explode instantly, as well as removing their velocity to prevent them from going flying?

src/main/java/dev/galacticraft/mod/content/entity/vehicle/RocketEntity.java
```diff
@@ -393,6 +407,176 @@ public void removePassenger(Entity entity) {
...
+            }
+
+            for (PrimedTnt primed : spawned) {
+                primed.setFuse(0);
+                primed.setDeltaMovement(Vec3.ZERO);
+
+                primed.tick();
+            }
+
+            if (spawned.isEmpty()) {
```
Roelymole: You can avoid creating a new list/one of the for loops using a simple flag variable in this sort of way:

```suggestion
            boolean spawned = false;
            for (PrimedTnt primed : after) {
                if (!before.contains(primed)) {
                    spawned = true;
                    primed.setFuse(0);
                    primed.setDeltaMovement(Vec3.ZERO);
                    primed.tick();
                }
            }

            if (!spawned) {
```

src/main/java/dev/galacticraft/mod/content/entity/vehicle/RocketEntity.java
```diff
@@ -393,6 +407,176 @@ public void removePassenger(Entity entity) {
...
+
+        final double scanRadius = 10.0;
+        List<PrimedTnt> before = level.getEntitiesOfClass(
+                PrimedTnt.class,
+                this.getBoundingBox().inflate(scanRadius)
+        );
+
+        BlockState oldState = level.getBlockState(pos);
+        BlockState payloadState = tntBlock.defaultBlockState();
+        level.setBlock(pos, payloadState, 3);
```
Roelymole: Using the constant `Block.UPDATE_ALL` is much clearer than using the value of 3 directly.

src/main/java/dev/galacticraft/mod/content/entity/vehicle/RocketEntity.java
```diff
@@ -393,6 +407,176 @@ public void removePassenger(Entity entity) {
...
+
+            if (spawned.isEmpty()) {
+                LivingEntity igniter = (this.getFirstPassenger() instanceof LivingEntity le) ? le : null;
+                PrimedTnt primed = new PrimedTnt(level, this.getX(), this.getY(), this.getZ(), igniter);
+                primed.setFuse(0);
+                level.addFreshEntity(primed);
+                primed.tick();
+            }
+        } finally {
+            level.setBlock(pos, oldState, 3);
```
Roelymole: The constant `Block.UPDATE_ALL` can be used here too.

src/main/java/dev/galacticraft/mod/content/entity/vehicle/RocketEntity.java
maxryan008: So the way this works is it gets the "before" list of tnt entities, then it runs the wasExploded() method which is how tnt blocks turn into entities, and it spawns a new piece of primed tnt, then we get the "after" list and whatever wasnt in the before list but is now in the after list can be assumed to have been spawned via the wasExploded() method. now that its been spawned however it has a velocity and a fuse. Similar to how minecraft tnt works. the wasExploded() method is essentially the same as applying redstone power to an explosive. Because it has a velocity it needs to be turned to zero so it looks like the rocket is where it exploded, and also the fuse can be anywhere from normal minecraft tnt length of 5 seconds all the way to 60 seconds or more for some modded tnt blocks. This meant when the rocket crashed it would simply summon a long fused tnt at the crash site which didnt look realistic IMO so the program is designed to set the fuse to zero to instantly explode (however technically some modded tnts are so big that the server and client desync and the server lags causing the client to see the rocket sort of just dissapear and then after a couple seconds the explosion happens, however thats just to do with how many block updates the server can handle at once).

PR #546, opened 2026-02-01
fix: overhaul bubble distributor GUI to prevent text overflowing

Roelymole:
- Split the status text onto two lines since "Status: Not Enough Energy" and "Status: Not Enough Oxygen" were overflowing.
- Increased the UI height slightly in order to fit everything in.
- Fixed the status being displayed as distributing when the machine was disabled.
- Made the green text darker to be more readable.
- General refactoring.


src/main/java/dev/galacticraft/mod/client/gui/screen/ingame/OxygenBubbleDistributorScreen.java
```diff
@@ -67,61 +74,79 @@ public OxygenBubbleDistributorScreen(OxygenBubbleDistributorMenu handler, Invent
...
-        textField.setValue(String.valueOf(this.menu.targetSize));
-
-        MachineStatus status = this.menu.state.getStatus();
-        graphics.drawString(this.font, Component.translatable(Translations.Ui.MACHINE_STATUS, status != null ? status.getText() : Component.empty()), this.leftPos + 60, this.topPos + 30, ChatFormatting.DARK_GRAY.getColor(), false);
 
-        this.textField.render(graphics, mouseX, mouseY, delta);
+        MutableComponent statusText = this.menu.state.getStatusText(this.menu.redstoneMode).copy();
+        if (statusText.getStyle().getColor() == TextColor.fromLegacyFormat(ChatFormatting.GREEN)) {
+            statusText.withStyle(ChatFormatting.DARK_GREEN);
+        }
```
marcus8448, line 140: Should we change the color to dark green everywhere? (update the status colours in GC/MachineLib)

Roelymole: I will try that out when I get a chance and see how it looks. At the very least I would do the same override for the other machines to be consistent.

PR #556, opened 2026-04-02
feat: rotated banner patterns

Roelymole:
Adds rotated versions of the vanilla banner patterns to be able to create flags with the pattern in the correct orientation.

<br>

These have been added to the corresponding banner pattern item tags, so the rotated version appears as an additional option in the loom.

<br>

This is a very simple solution, but trying to do this dynamically for arbitrary banner patterns probably wouldn't look good anyway.


src/main/java/dev/galacticraft/mod/data/GCTranslationProvider.java
```diff
@@ -828,6 +812,39 @@ protected void generateEntityTranslations() {
         this.entity(GCEntityTypes.THROWABLE_METEOR_CHUNK, "Meteor Chunk");
     }
 
+    protected void generateBannerPatternTranslations() {
```
JustAPotota, line 815: You could avoid hardcoding all the color names here by iterating over `DyeColor`s and getting the proper name with `normalizeName(color.getName())` like the dyed pipes do. Though ideally we'd have a function for making color set translations.

JustAPotota (changes requested):
I gave a quick test of all the rotated patterns and they look good to me! I think "sideways" would sound a little better for the pattern names, but that's a tiny nitpick.

PR #561, opened 2026-04-06
Add heavy aluminum wires and walkway

gregorixG:
It adds full support for Heavy Aluminum Wire and Heavy Sealable Aluminum Wire, including block registration, recipes, loot, tags, creative tab entries, translations, model generation, and legacy texture integration. It also fixes heavy-wire behavior so tier-2 wires connect and function like real power wires instead of rendering as inert blobs. Transfer limits now come from config, and tier-2 wires are correctly exposed through the energy lookup system so they connect to the same machines and networks as expected.

It also adds a complete Heavy Wire Walkway block, not just a translation or placeholder. That includes registration, block entity support, recipes, tags, loot, render-layer setup, blockstate/model generation, English and Hungarian naming, and proper heavy-wire radius/collision behavior. Follow-up rendering fixes addressed the heavy walkway’s visual issues: z-fighting, broken top joins, and the mismatch between the walkway wire and adjacent heavy wires. The final result uses dedicated heavy walkway platform geometry so the heavy wire remains visually continuous with normal heavy wire while still fitting cleanly inside the walkway.


src/main/java/dev/galacticraft/mod/client/model/WalkwayCenterModel.java
Roelymole, line 1: The indentation for some of your changes in this file seems to have got messed up.

src/main/java/dev/galacticraft/mod/data/model/GCModelProvider.java
```diff
@@ -483,7 +486,10 @@ private static void createActiveMachine(BlockModelGenerators generator, Block bl
     }
 
     private static void createWalkway(BlockModelGenerators generator, Block walkway, ResourceLocation pipeModel, ResourceLocation centerModel) {
-        ResourceLocation walkwayPlatform = Constant.id("block/walkway");
+                createWalkway(generator, walkway, pipeModel, centerModel, Constant.id("block/walkway"));
+        }
+
+        private static void createWalkway(BlockModelGenerators generator, Block walkway, ResourceLocation pipeModel, ResourceLocation centerModel, ResourceLocation walkwayPlatform) {
```
Roelymole: These lines appear to be indented by 4 spaces too many.

src/main/java/dev/galacticraft/mod/data/GCBlockLootTableProvider.java
```diff
@@ -201,6 +202,7 @@ public void generate() {
         }
 
         this.dropSelf(GCBlocks.ALUMINUM_WIRE);
+    this.dropSelf(GCBlocks.HEAVY_ALUMINUM_WIRE);
```
Roelymole: The indentation of this line isn't quite right.

src/main/java/dev/galacticraft/mod/data/tag/GCBlockTagProvider.java
```diff
@@ -596,12 +596,14 @@ protected void addTags(HolderLookup.Provider provider) {
...
                         GCBlocks.WALKWAY,
                         GCBlocks.WIRE_WALKWAY,
+                        GCBlocks.HEAVY_WIRE_WALKWAY,
                         GCBlocks.FLUID_PIPE_WALKWAY,
                         GCBlocks.CRYOGENIC_CHAMBER,
                         GCBlocks.CRYOGENIC_CHAMBER_PART,
                         GCBlocks.SOLAR_PANEL_PART,
                         GCBlocks.AIR_LOCK_FRAME,
                         GCBlocks.AIR_LOCK_CONTROLLER,
+                        GCBlocks.HEAVY_ALUMINUM_WIRE,
```
Roelymole, line 607: `GCBlocks.ALUMINUM_WIRE` should be added to `BlockTags.MINEABLE_WITH_PICKAXE` as well.

src/main/java/dev/galacticraft/mod/data/recipes/GCMachineRecipes.java
```diff
@@ -370,6 +381,14 @@ public void buildRecipes(RecipeOutput output) {
                 .emiDefaultRecipe(true)
                 .save(output);
 
+        GCShapedRecipeBuilder.crafting(RecipeCategory.DECORATIONS, GCBlocks.HEAVY_SEALABLE_ALUMINUM_WIRE, 6)
```
Roelymole: I'm aware that this is copying how many items the recipe for the standard sealable aluminum wire produces, but 6 seems like a bit much for both.

The most likely explanation is that it was copied from the aluminum ingot and wool recipe and not changed to a more suitable value.

I think Galacticraft Legacy only gave 1 item, so that might be a better value to go with for now, but I would be open to rethinking this whole recipe.

src/main/java/dev/galacticraft/mod/data/recipes/GCMachineRecipes.java
```diff
@@ -362,6 +362,17 @@ public void buildRecipes(RecipeOutput output) {
                 .emiDefaultRecipe(true)
                 .save(output);
 
+        GCShapedRecipeBuilder.crafting(RecipeCategory.DECORATIONS, GCBlocks.HEAVY_ALUMINUM_WIRE, 6)
```
Roelymole: This recipe should only produce 1 heavy aluminum wire, not 6.

src/main/java/dev/galacticraft/mod/data/model/GCModelProvider.java
```diff
@@ -692,13 +698,15 @@ public void generateItemModels(ItemModelGenerators generator) {
         }
 
         GCModelTemplates.PIPE_INVENTORY.create(ModelLocationUtils.getModelLocation(GCBlocks.ALUMINUM_WIRE.asItem()), TextureMapping.layer0(GCBlocks.ALUMINUM_WIRE), generator.output);
+        generator.generateFlatItem(GCBlocks.HEAVY_ALUMINUM_WIRE.asItem(), ModelTemplates.FLAT_ITEM);
```
Roelymole: The item model for the heavy aluminum wire should be similar to the item model for the standard aluminum wire (a 3D model, not a flat item).

src/main/java/dev/galacticraft/mod/client/model/WalkwayCenterModel.java
```diff
@@ -128,38 +147,38 @@ public Baked(Function<Material, TextureAtlasSprite> textureGetter, ResourceLocat
...
-                    .square(Direction.DOWN, 0.375f, 0.0625f, 0.625f, 0.375f, 0.375f)
+                    .square(Direction.DOWN, min, edgeMin, max, min, min)
                     .uv(0, 0, 10)
                     .uv(1, 0, 15)
                     .uv(2, 4, 15)
                     .uv(3, 4, 10)
                     .spriteBake(this.sprite, MutableQuadView.BAKE_NORMALIZED & MutableQuadView.BAKE_LOCK_UV)
                     .color(-1, -1, -1, -1).emit();
-            emitter.square(Direction.NORTH, 0.375f, 0.375f, 0.625f, 0.625f, 0.0625f)
+            emitter.square(Direction.NORTH, min, min, max, max, edgeMin)
```
Roelymole: There is a mix of `emitter` being on its own line and being on the same line as `.square(...)` which is not ideal. This seems like a good opportunity to make that consistent one way or the other.

src/main/java/dev/galacticraft/mod/data/recipes/GCMachineRecipes.java
```diff
@@ -362,6 +362,17 @@ public void buildRecipes(RecipeOutput output) {
...
+                .define('A', GCItemTags.ALUMINUM_INGOTS)
+                .define('W', GCBlocks.ALUMINUM_WIRE)
+                .define('L', ItemTags.WOOL)
+                .pattern(" A ")
+                .pattern(" W ")
+                .pattern(" L ")
```
Roelymole: It might be best to remove these spaces, although the recipe seems to work.

Could you also add a version of this recipe that is mirrored vertically.

src/main/java/dev/galacticraft/mod/content/GCBlocks.java
```diff
@@ -314,6 +317,7 @@ public class GCBlocks {
     public static final Block WALKWAY = BLOCKS.registerWithItem(Constant.Block.WALKWAY, new WalkwayBlock(BlockBehaviour.Properties.of().mapColor(MapColor.METAL).strength(5.0f, 5.0f).sound(SoundType.METAL).requiresCorrectToolForDrops().noOcclusion()));
     public static final Block FLUID_PIPE_WALKWAY = BLOCKS.registerWithItem(Constant.Block.FLUID_PIPE_WALKWAY, new GlassFluidPipeWalkwayBlock(BlockBehaviour.Properties.ofFullCopy(GLASS_FLUID_PIPE), PipeColor.CLEAR));
     public static final Block WIRE_WALKWAY = BLOCKS.registerWithItem(Constant.Block.WIRE_WALKWAY, new WireWalkwayBlock(BlockBehaviour.Properties.ofFullCopy(ALUMINUM_WIRE)));
+    public static final Block HEAVY_WIRE_WALKWAY = BLOCKS.registerWithItem(Constant.Block.HEAVY_WIRE_WALKWAY, new HeavyWireWalkwayBlock(BlockBehaviour.Properties.ofFullCopy(HEAVY_ALUMINUM_WIRE)));
```
Roelymole: I think the pipe/wire walkways should inherit most/all of their properties from the normal walkway, rather than the pipes/wires.

src/main/java/dev/galacticraft/mod/Constant.java
Roelymole: There isn't really a natural place that the heavy wires fit in here, so it might be nice to put them all into a more sensible order.

Roelymole (changes requested):
This is looking really good! I have a few minor issues, mostly to do with the indentation and recipes.

I'll review the model side of things again when you have pushed those latest changes.

.gitignore
```diff
@@ -222,3 +222,6 @@ src/main/generated/.cache/
 
 # Python 
 .github/scripts/__pycache__
+
+# Local-only localization override
+src/main/resources/assets/galacticraft/lang/hu_hu.json
```
Roelymole: Since this is a personal thing it would be better to add it to `.git/info/exclude` which is basically a local `.gitignore`.

src/main/java/dev/galacticraft/mod/client/model/PipeBakedModel.java
Roelymole, line 1: You have fixed the indentation in the other files, but somehow messed it up in this one when it was fine before...

src/main/java/dev/galacticraft/mod/client/model/PipeBakedModel.java
```diff
@@ -239,20 +254,101 @@ public void emitBlockQuads(BlockAndTintGetter getter, BlockState blockState, Blo
...
             }
         }
+
+    }
+
+    private Mesh buildUpMesh(float minY, float maxY) {
+        MeshBuilder meshBuilder = RendererAccess.INSTANCE.getRenderer().meshBuilder();
+        QuadEmitter emitter = meshBuilder.getEmitter();
+
+        for (Direction direction : new Direction[]{Direction.NORTH, Direction.SOUTH, Direction.EAST, Direction.WEST}) {
```
Roelymole: This can be simplified. I think there are a few other places where this can be done too.

```suggestion
        for (Direction direction : Direction.Plane.HORIZONTAL) {
```

src/main/java/dev/galacticraft/mod/data/model/GCModelProvider.java
```diff
@@ -692,13 +698,15 @@ public void generateItemModels(ItemModelGenerators generator) {
         }
 
         GCModelTemplates.PIPE_INVENTORY.create(ModelLocationUtils.getModelLocation(GCBlocks.ALUMINUM_WIRE.asItem()), TextureMapping.layer0(GCBlocks.ALUMINUM_WIRE), generator.output);
+        GCModelTemplates.PIPE_INVENTORY.create(ModelLocationUtils.getModelLocation(GCBlocks.HEAVY_ALUMINUM_WIRE.asItem()), TextureMapping.layer0(GCBlocks.HEAVY_ALUMINUM_WIRE), generator.output);
```
Roelymole: We will need to use a different template for the heavy wires.

src/main/java/dev/galacticraft/mod/data/model/GCModelProvider.java
```diff
@@ -692,13 +698,15 @@ public void generateItemModels(ItemModelGenerators generator) {
...
         for (Block pipe : GCBlocks.GLASS_FLUID_PIPES.values()) {
             GCModelTemplates.PIPE_INVENTORY.create(ModelLocationUtils.getModelLocation(pipe.asItem()), TextureMapping.layer0(pipe), generator.output);
         }
 
         GCModelTemplates.WALKWAY_INVENTORY.create(ModelLocationUtils.getModelLocation(GCBlocks.WALKWAY.asItem()), TextureMapping.defaultTexture(Constant.id("block/walkway_connector")), generator.output);
         GCModelTemplates.WALKWAY_INVENTORY.create(ModelLocationUtils.getModelLocation(GCBlocks.FLUID_PIPE_WALKWAY.asItem()), TextureMapping.defaultTexture(Constant.id("block/glass_fluid_pipe")), generator.output);
         GCModelTemplates.WALKWAY_INVENTORY.create(ModelLocationUtils.getModelLocation(GCBlocks.WIRE_WALKWAY.asItem()), TextureMapping.defaultTexture(Constant.id("block/aluminum_wire")), generator.output);
+        GCModelTemplates.WALKWAY_INVENTORY.create(ModelLocationUtils.getModelLocation(GCBlocks.HEAVY_WIRE_WALKWAY.asItem()), TextureMapping.defaultTexture(Constant.id("block/heavy_aluminum_wire")), generator.output);
```
Roelymole: We will need to use a different template for the heavy wire walkway too, or at least the wire part.

src/main/java/dev/galacticraft/mod/client/model/PipeBakedModel.java
```diff
@@ -54,25 +55,39 @@
...
+
+    public PipeBakedModel(Function<Material, TextureAtlasSprite> textureGetter, ResourceLocation texture, float radius, float yOffset) {
         this.sprite = textureGetter.apply(new Material(InventoryMenu.BLOCK_ATLAS, texture));
         this.meshes = new EnumMap<>(Direction.class);
         this.radius = radius;
+        this.yOffset = yOffset;
+        this.topConnectionMesh = yOffset < 0.0f ? this.buildUpMesh(0.5f + radius + yOffset, 1.0f) : null;
+
+        float crossSectionWidth = this.capPixelWidth();
+        float positiveArmStart = 10.0f;
```
Roelymole: This value needs to change based on the radius of the wire.

Also, I can see what you mean by `positiveArmStart`, but I think `centerStart` and `centerEnd` might capture the same idea more clearly.

```suggestion
        float centerStart = (0.5f - this.radius) * 16.0f;
        float centerEnd = (0.5f + this.radius) * 16.0f;
```

`centerStart` should take the value of `6.0f` for the standard wires and `5.0f` for the heavy wires and `centerEnd` should take the value of `10.0f` for the standard wires and `11.0f` for the heavy wires.

src/main/java/dev/galacticraft/mod/client/model/PipeBakedModel.java
```diff
@@ -83,8 +98,8 @@ public PipeBakedModel(Function<Material, TextureAtlasSprite> textureGetter, Reso
                     .square(direction, 0.5f-radius, 0.5f+radius, 0.5f+radius, 1.0f, 0.5f-radius)
                     .uv(0, 0, 0)
                     .uv(1, 0, 6)
-                    .uv(2, 4, 6)
```
Roelymole: Following on from my comment above, you would need to replace the values of 6 in the third argument to `uv` with `centerStart` (here and elsewhere).

```suggestion
                    .uv(1, 0, centerStart)
                    .uv(2, crossSectionWidth, centerStart)
```

src/main/java/dev/galacticraft/mod/client/model/PipeBakedModel.java
```diff
@@ -239,20 +254,101 @@ public void emitBlockQuads(BlockAndTintGetter getter, BlockState blockState, Blo
...
+    private float capU(float value) {
+        return (value - (0.5f - this.radius)) * (this.capPixelWidth() / (this.radius * 2.0f));
     }
+
+    private float capV(float value) {
+        return 6.0f + (value - (0.5f - this.radius)) * (this.capPixelWidth() / (this.radius * 2.0f));
+    }
+
+    private float capPixelWidth() {
+        return this.radius * 32.0f;
```
Roelymole: Based on this definition `this.capPixelWidth() / (this.radius * 2.0f)` is always `16.0f`, so the `capU` and `capV` methods can be simplified.

src/main/java/dev/galacticraft/mod/client/model/WalkwayCenterModel.java
```diff
@@ -69,231 +87,236 @@ public WalkwayCenterModel(ResourceLocation texture) {
...
+            return (value - (0.5f - this.radius)) * (this.capPixelWidth() / (this.radius * 2.0f));
+        }
+
+        private float capV(float value) {
+            return 6.0f + (value - (0.5f - this.radius)) * (this.capPixelWidth() / (this.radius * 2.0f));
+        }
+
+        private float capPixelWidth() {
+            return this.radius * 32.0f;
         }
```
Roelymole, line 309: These would need to be updated to match their counterparts in `PipeBakedModel`.

src/main/java/dev/galacticraft/mod/client/model/PipeBakedModel.java
```diff
@@ -239,20 +260,112 @@ public void emitBlockQuads(BlockAndTintGetter getter, BlockState blockState, Blo
...
+                .spriteBake(this.sprite, MutableQuadView.BAKE_NORMALIZED & MutableQuadView.BAKE_LOCK_UV)
+                .color(-1, -1, -1, -1).emit();
+    }
+
+    private float capU(float value) {
+        return (value - (0.5f - this.radius)) * 16.0f;
+    }
+
+    private float capV(float value) {
+        return this.centerStart() + (value - (0.5f - this.radius)) * 16.0f;
```
Roelymole: This can be simplified to just `value * 16.0f`.

src/main/java/dev/galacticraft/mod/content/block/special/walkway/AbstractWalkwayBlock.java
```diff
@@ -43,28 +43,39 @@
...
+    static Map<Pair<Integer, Direction>, VoxelShape> getShapes(float pipeRadius) {
+        return SHAPES.computeIfAbsent(pipeRadius, radius -> Util.make(new HashMap<>(), map -> {
+            VoxelShape[] pipeShapes = PipeShapedBlock.makeShapes(radius);
+
+            for (int pipeAabb = 0; pipeAabb < Math.pow(2, 6); pipeAabb++) {
+                for (Direction platformDirection : Direction.values()) {
+                    map.put(Pair.of(pipeAabb, platformDirection), Shapes.or(
+                            pipeShapes[pipeAabb],
+                            ConnectingBlockUtil.WALKWAY_SHAPES.get(platformDirection)
+                    ));
```
Roelymole, line 76: This is a very minor detail, but the hitbox for the wire going through the center of the walkway is only present when it is connected to another wire.

Image

Image<br>

You should just be able to add another shape to this `Shapes.or(...)` to ensure the hitbox is consistent.

src/main/java/dev/galacticraft/mod/client/model/PipeBakedModel.java
```diff
@@ -239,20 +260,112 @@ public void emitBlockQuads(BlockAndTintGetter getter, BlockState blockState, Blo
...
+        float minOuter = 0.5f - this.radius;
+        float maxOuter = 0.5f + this.radius;
+        float minInner = 0.5f - neighborRadius;
+        float maxInner = 0.5f + neighborRadius;
+        float centerInset = Math.min(REDUCER_CENTER_THICKNESS, this.radius - neighborRadius);
+
+        this.emitCapQuad(emitter, direction, minInner, minInner, maxInner, maxInner, EDGE_FACE_DEPTH);
+        if (centerInset > 0.0f) {
+            this.emitCapQuad(emitter, direction, minInner, minInner, maxInner, maxInner, EDGE_FACE_DEPTH + centerInset);
+        }
```
Roelymole: Correct me if I'm wrong, but I don't think we need this as all it does is render an extra face inside the wire.

Image

src/main/java/dev/galacticraft/mod/client/model/WalkwayCenterModel.java
```diff
@@ -69,231 +87,246 @@ public WalkwayCenterModel(ResourceLocation texture) {
...
                     .color(-1, -1, -1, -1).emit();
-            this.meshes.put(Direction.WEST, meshBuilder.build());
+        }
+
+        private float capU(float value) {
+            return (value - (0.5f - this.radius)) * 16.0f;
+        }
+
+        private float capV(float value) {
+            return this.centerStart() + (value - (0.5f - this.radius)) * 16.0f;
```
Roelymole: This can also be simplified to `value * 16.0f`.

Roelymole (approved):
A few last things and then we should be good to go!

PR #562, opened 2026-04-07
refactor: ServerLevel instances are now labeled as "level" instead of "world"

TakeABao23:
I found that our naming convention for `ServerLevel` instances was inconsistent, so I fixed it for all machine blocks. All instances of `ServerLevel` in machine files are now labeled as `level`.


TakeABao23:
Edits outdated code. Will redo this more cleanly and with updated code.

PR #568, opened 2026-04-17
Withdrawn

Youraveragebub:


src/main/java/dev/galacticraft/mod/misc/footprint/ServerFootprintManager.java
```diff
@@ -33,12 +33,23 @@
...
+    @Override
+    public void onFootprintAdded(Level level, long packedPos, Footprint footprint) {
+        if (footprint.type == FootprintType.HUMAN) {
+            return;
+        }
+
+        PlayerLookup.tracking((ServerLevel) level, new ChunkPos(packedPos)).forEach(player ->
+                ServerPlayNetworking.send(player, new FootprintPacket(packedPos, List.of(footprint)))
+        );
+    }
```
Roelymole: If something like this is required for pets, then it is probably needed for other players.

PR #543, opened 2026-01-27
feat: add sounds for all powered machines

TakeABao23:
I added sound files, factored them, and added functionality to play them whenever a machine is powered. All machines which run on electricity have these sounds now.

PR #572, opened 2026-04-29
feat: add dynamic sound system

TakeABao23:
I have added a dynamic sound system and some new sound instance classes in order to facilitate the further creation of custom sounds for GC5. Every machine in the game has this functionality, though only electric-powered machines have sounds assigned to them.

(**Note**: This is a re-upload of an [older pull request.](https://github.com/TeamGalacticraft/Galacticraft/pull/543) Since the original pull request, I moved my workspace to a separate branch for cleanliness and implemented all the requested changes from the original.)

# Sound System Features:
- A sound manager that handles custom looping GC5 sounds
- A generic machine sound class
- State change tracking
- State change sound modulation (fading in and out)

# Machine Sounds Features:
- Unique sound that plays when the machine is powered
- Unique sound that plays when the machine is active
- Circuit scritch sounds play randomly as the Circuit Fabricator runs
- Separate (currently empty) map of sounds for coal-powered machines

# Feedback sought:
- Volume: how loud are single machines? Groups of machines?
- Quality: do they sound like what they're supposed to?
- Performance: does this system hurt performance in any way?

# Known issues:
- Upon relogging, sounds won't play until the machine status updates

# Future Plans:
- Add more unique sounds (particularly for coal-powered machines)
- possible unique sound behaviors for different states?


TakeABao23:
It is finished! The sound system is feature complete!

src/main/java/dev/galacticraft/mod/events/ClientEventHandler.java
```diff
@@ -33,9 +33,11 @@
 import dev.galacticraft.mod.client.render.FootprintRenderer;
 import dev.galacticraft.mod.client.render.GCBlockOutlineRenderer;
 import dev.galacticraft.mod.client.render.dimension.SatelliteSkyRenderer;
+import dev.galacticraft.mod.client.sounds.GCSoundManager;
 import dev.galacticraft.mod.client.sounds.RocketSound;
 import dev.galacticraft.mod.content.entity.vehicle.RocketEntity;
 import dev.galacticraft.mod.misc.footprint.FootprintManager;
+import dev.galacticraft.machinelib.client.api.event.MachineStatusEvents;
```
Roelymole: `dev.galacticraft.machinelib` comes before `dev.galacticraft.mod` alphabetically.

src/main/java/dev/galacticraft/mod/client/sounds/GCSoundManager.java
```diff
@@ -0,0 +1,133 @@
...
+                break;
+            case OUTPUT_FULL:
+                manager.play(new IdleMachineSound(machine,newSound, manager));
+                break;
+            case PARTIALLY_WORKING:
+                manager.play(new ActiveMachineSound(machine,newSound, manager));
+                break;
+            case WORKING:
+                manager.play(new ActiveMachineSound(machine,newSound, manager));
+                break;
```
Roelymole: This can be made more concise. It also highlights which cases are handled in the same way.

```suggestion
            case MISSING_FLUIDS:
            case MISSING_ITEMS:
            case MISSING_RESOURCE:
            case OTHER:
            case OUTPUT_FULL:
                manager.play(new IdleMachineSound(machine, newSound, manager));
                break;
            case PARTIALLY_WORKING:
            case WORKING:
                manager.play(new ActiveMachineSound(machine, newSound, manager));
                break;
```

src/main/java/dev/galacticraft/mod/client/sounds/TransitionState.java
Roelymole, line 1: It might be better to define this enum within the `GCSound` class, since that is the only place it is being used. The `transitionState` attribute is protected with no getters or setters, so the enum could be protected too.

src/main/java/dev/galacticraft/mod/events/MachineEvents.java
Roelymole, line 1: If you decided you didn't end up needing this class, please delete it.

src/main/resources/assets/galacticraft/sounds.json
```diff
@@ -180,6 +180,37 @@
       }
     ]
   },
+  "machine.circuitscritch": {
+  "sounds": [
+    "galacticraft:machine/circuitscritch1",
+    "galacticraft:machine/circuitscritch3",
+    "galacticraft:machine/circuitscritch4"
```
Roelymole: Is there a reason why `"galacticraft:machine/circuitscritch2"` has been omitted?

src/main/java/dev/galacticraft/mod/client/sounds/GCSoundManager.java
```diff
@@ -0,0 +1,133 @@
...
+import net.minecraft.sounds.SoundEvent;
+import net.minecraft.world.level.block.entity.BlockEntity;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+
+import dev.galacticraft.machinelib.api.block.entity.MachineBlockEntity;
+import dev.galacticraft.machinelib.api.machine.MachineStatus;
```
Roelymole: The convention we have been using throughout the codebase is to order the imports like follows:

```suggestion
import dev.galacticraft.machinelib.api.block.entity.MachineBlockEntity;
import dev.galacticraft.machinelib.api.machine.MachineStatus;
import net.minecraft.client.Minecraft;
import net.minecraft.client.player.LocalPlayer;
import net.minecraft.core.BlockPos;
import net.minecraft.sounds.SoundEvent;
import net.minecraft.world.level.block.entity.BlockEntity;

import java.util.ArrayList;
import java.util.List;
import java.util.Objects;
import java.util.Optional;
```

src/main/java/dev/galacticraft/mod/client/sounds/GCSoundMap.java
```diff
@@ -0,0 +1,99 @@
...
+package dev.galacticraft.mod.client.sounds;
+
+import dev.galacticraft.machinelib.api.machine.MachineStatus;
+import dev.galacticraft.machinelib.api.machine.MachineStatuses;
+import dev.galacticraft.mod.content.GCSounds;
+import dev.galacticraft.mod.machine.GCMachineStatuses;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import net.minecraft.sounds.SoundEvent;
```
Roelymole, line 31: Again, the convention we have been using to arrange the imports is like follows:

```suggestion
import dev.galacticraft.machinelib.api.machine.MachineStatus;
import dev.galacticraft.machinelib.api.machine.MachineStatuses;
import dev.galacticraft.mod.content.GCSounds;
import dev.galacticraft.mod.machine.GCMachineStatuses;
import net.minecraft.sounds.SoundEvent;

import java.util.Collections;
import java.util.HashMap;
import java.util.Map;
```

src/main/java/dev/galacticraft/mod/client/sounds/GCSound.java
```diff
@@ -0,0 +1,107 @@
...
+            case ENDING:
+                this.transitionTick++;
+                if (this.transitionTick > this.endTransitionTicks) {
+                    this.callback.onFinished(this);
+                }
+                break;
+            case RUNNING:
+                break;
+            }
+        }
```
Roelymole: These closing braces are indented by 4 spaces too many each.

src/main/java/dev/galacticraft/mod/client/sounds/MachineSound.java
```diff
@@ -0,0 +1,53 @@
...
+import dev.galacticraft.machinelib.api.block.entity.MachineBlockEntity;
+import dev.galacticraft.machinelib.api.machine.MachineStatus;
+import net.minecraft.sounds.SoundEvent;
+import net.minecraft.sounds.SoundSource;
+
+
+public abstract class MachineSound extends GCSound {
+    protected final MachineBlockEntity machine;
+    GCSound newsound;
+    MachineStatus status;
```
Roelymole: As far as I can see these are unused.

src/main/java/dev/galacticraft/mod/client/sounds/MachineSound.java
Roelymole, line 1: This class doesn't have any abstract methods, so it doesn't really need to be abstract. Then the `ActiveMachineSound` and `IdleMachineSound` classes can be replaced, since all they are really doing is providing default values of arguments.

src/main/java/dev/galacticraft/mod/client/sounds/GCSoundManager.java
```diff
@@ -0,0 +1,133 @@
...
+        return Optional.empty();
+    }
+
+    public static void onStatusChanged(Minecraft minecraft, LocalPlayer player, BlockPos pos, MachineStatus status, MachineStatus oldStatus) {
+        MachineBlockEntity machine = (MachineBlockEntity) minecraft.level.getBlockEntity(pos);
+        GCSoundManager manager = GCSoundManager.getInstance();
+        SoundEvent newSound = GCSoundMap.GC_SOUND_MAP.get(status);
+
+        // Stop old sound (if there is one)
+        manager.getSoundFromEntity(machine, oldStatus).ifPresent(oldSound->{oldSound.end();});
```
Roelymole: For a single statement the curly brackets and semi-colon can be omitted.

```suggestion
        manager.getSoundFromEntity(machine, oldStatus).ifPresent(oldSound -> oldSound.end());
```

src/main/java/dev/galacticraft/mod/client/sounds/GCSound.java
```diff
@@ -0,0 +1,107 @@
...
+import net.minecraft.client.resources.sounds.SoundInstance;
+import net.minecraft.sounds.SoundEvent;
+import net.minecraft.sounds.SoundSource;
+import net.minecraft.util.Mth;
+import net.minecraft.world.level.block.entity.BlockEntity;
+
+public abstract class GCSound extends AbstractTickableSoundInstance {
+    protected final BlockEntity entity;
+    protected final SoundCallback callback;
+    protected TransitionState transitionState;
```
Roelymole: This attribute could probably be initialised to `TransitionState.STARTING` here rather than in the constructor, since I can't forsee initialising it to any of the other values making sense.

src/main/java/dev/galacticraft/mod/client/sounds/MachineSound.java
```diff
@@ -0,0 +1,53 @@
...
+        if (this.entity instanceof MachineBlockEntity blockEntity && blockEntity.isRemoved()) {
+            this.end();
+        }
+        super.tick();
+        this.modulateSoundforTransition();
+    }
+
+    public MachineStatus getStatus() {
+        return this.machine.getState().getStatus();
+    }
```
Roelymole: This isn't being used. I'm also not sure if the value returned by the `getState()` method on the client-side can be trusted.

src/main/java/dev/galacticraft/mod/client/sounds/MachineSound.java
```diff
@@ -0,0 +1,53 @@
...
+package dev.galacticraft.mod.client.sounds;
+
+import dev.galacticraft.machinelib.api.block.entity.MachineBlockEntity;
+import dev.galacticraft.machinelib.api.machine.MachineStatus;
+import net.minecraft.sounds.SoundEvent;
+import net.minecraft.sounds.SoundSource;
+
+
+public abstract class MachineSound extends GCSound {
+    protected final MachineBlockEntity machine;
```
Roelymole: This is a bit of a duplication of the `entity` attribute in `GCSound`. I'm not sure if the extra level of abstraction is serving its purpose. Perhaps `GCSound` and `MachineSound` should be combined together (I would keep the name `MachineSound`).

src/main/java/dev/galacticraft/mod/content/block/entity/machine/CircuitFabricatorBlockEntity.java
```diff
@@ -110,6 +113,13 @@ public CircuitFabricatorBlockEntity(BlockPos pos, BlockState state) {
         super(GCBlockEntityTypes.CIRCUIT_FABRICATOR, pos, state, GCRecipes.FABRICATION_TYPE, SPEC);
     }
 
+    public void workingSounds(MachineStatus status) {
+        RandomSource randomSource=RandomSource.create();
+        if (randomSource.nextDouble() < 0.05 && status==GCMachineStatuses.FABRICATING) {
```
Roelymole: You can use `level.random` instead of creating a new `RandomSource` (like you are for the pitch argument already).

```suggestion
        if (level.random.nextDouble() < 0.05 && status == GCMachineStatuses.FABRICATING) {
```

src/main/java/dev/galacticraft/mod/client/sounds/GCSoundMap.java
Roelymole, line 1: I am not particularly a fan of this approach and there seems to be quite a bit of redundancy here too. I need a bit more time to think about the best way to do this and hopefully things will be a bit clearer after some of the other changes have been made too.

Roelymole (changes requested):
There are mostly refactoring changes I would like to see.

I just added a few things (mostly regarding whitespace) to the [code style](https://github.com/TeamGalacticraft/Galacticraft?tab=contributing-ov-file#code-style) section of the contributing guidelines.

Please could you make sure you are following that style more consistently.

src/main/java/dev/galacticraft/mod/client/sounds/GCSoundMap.java
TakeABao23: Yeah I'm not a huge fan of it either. The nice part is that we could have a custom sound for every single machine state possible, but since we don't actually have that right now I'd love ideas on how to only include the relevant ones.

TakeABao23: Perhaps we could have some default value based on status type?

src/main/java/dev/galacticraft/mod/client/sounds/MachineSound.java
```diff
@@ -0,0 +1,116 @@
...
+        float normTick = switch (transitionState) {
+            case STARTING -> (float) this.transitionTick/this.startTransitionTicks;
+            case ENDING -> 1.0F - ((float) this.transitionTick/this.endTransitionTicks);
+            case RUNNING -> 1.0F;
+        };
+        this.volume = Mth.lerp(normTick, 0.0F, this.maxVolume);
+    }
+
+    public enum TransitionState {
+    STARTING, RUNNING, ENDING
```
Roelymole: This should be indented 4 spaces more.

src/main/java/dev/galacticraft/mod/client/sounds/MachineSound.java
```diff
@@ -0,0 +1,116 @@
...
+
+    @Override
+    public boolean canStartSilent() {
+        return true;
+    }
+
+    protected void modulateSoundforTransition() {
+        float normTick = switch (transitionState) {
+            case STARTING -> (float) this.transitionTick/this.startTransitionTicks;
+            case ENDING -> 1.0F - ((float) this.transitionTick/this.endTransitionTicks);
```
Roelymole: Please could you put a space either side of the division operators.

src/main/java/dev/galacticraft/mod/client/sounds/MachineSound.java
```diff
@@ -0,0 +1,116 @@
...
+    protected final int startTransitionTicks, endTransitionTicks;
+    protected int transitionTick = 0;
+    protected final float maxVolume;
+    protected final SoundEvent event;
+
+    public MachineSound(MachineBlockEntity machine, SoundEvent event, SoundCallback callback, float maxVolume) {
+        super(event, SoundSource.BLOCKS, SoundInstance.createUnseededRandom());
+        // references to other objects
+        this.machine = machine;
+        this.event = event;
```
Roelymole, line 46: This value is not used again, so it doesn't need to be stored.

src/main/java/dev/galacticraft/mod/client/sounds/MachineSound.java
```diff
@@ -0,0 +1,116 @@
...
+        this.setPosition();
+    }
+
+    @Override
+    public void tick() {
+        if (this.machine instanceof MachineBlockEntity blockEntity && blockEntity.isRemoved()) {
+            this.end();
+        }
+        if (this.machine == null) {
+            this.callback.onFinished(this);
```
Roelymole, line 66: Is it intentional that the rest of the method continues, or should there be an early return here?

src/main/resources/assets/galacticraft/sounds.json
Roelymole, line 1: The indentation of the changes in this file is a bit off.

src/main/java/dev/galacticraft/mod/client/sounds/MachineSound.java
TakeABao23: oops missed that one sorry!

src/main/java/dev/galacticraft/mod/client/sounds/MachineSound.java
TakeABao23: It's actually used in GCSoundManager, line 89, due to the AbstractSoundInstance's soundEvent being inaccessible as a SoundEvent. Ideas for how I could get the soundEvent without needing to store the value?

Roelymole: I missed that one.

src/main/java/dev/galacticraft/mod/client/sounds/GCSoundManager.java
```diff
@@ -0,0 +1,122 @@
...
+            if (activeSound.getLocation().equals(soundEvent.getLocation())) {
+                return Optional.of(activeSound);
+            }
+        }
+        return Optional.empty();
+    }
+
+    public Optional<MachineSound> getSoundFromEntity(BlockEntity entity, MachineStatus status) {
+        for (var activeSound : this.activeSounds) {
+            if (activeSound.machine == entity && Objects.equals(GCSoundMap.GC_SOUND_MAP.get(status),activeSound.event)) {
```
Roelymole: Please could you put spaces after the commas here too.

src/main/java/dev/galacticraft/mod/client/sounds/MachineSound.java
```diff
@@ -0,0 +1,117 @@
...
+            case STARTING -> (float) this.transitionTick / this.startTransitionTicks;
+            case ENDING -> 1.0F - ((float) this.transitionTick / this.endTransitionTicks);
+            case RUNNING -> 1.0F;
+        };
+        this.volume = Mth.lerp(normTick, 0.0F, this.maxVolume);
+    }
+
+        public enum TransitionState {
+        STARTING, RUNNING, ENDING
+        }
```
Roelymole: I meant like this.

```suggestion
    public enum TransitionState {
        STARTING, RUNNING, ENDING
    }

```

src/main/resources/assets/galacticraft/sounds.json
```diff
@@ -180,6 +180,37 @@
...
     ]
   },
+  "machine.circuitscritch": {
+    "sounds": [
+      "galacticraft:machine/circuitscritch1",
+      "galacticraft:machine/circuitscritch2",
+      "galacticraft:machine/circuitscritch3"
+    ]
+  },
+    "machine.machinebuzz": {
```
Roelymole:
```suggestion
  "machine.machinebuzz": {
```

src/main/resources/assets/galacticraft/sounds.json
```diff
@@ -180,6 +180,37 @@
...
+    "sounds": [
+      {
+        "name": "galacticraft:machine/machinebuzz",
+        "stream": false
+      }
+    ]
+  },
+    "machine.machinehum": {
```
Roelymole:
```suggestion
  "machine.machinehum": {
```

src/main/resources/assets/galacticraft/sounds.json
```diff
@@ -180,6 +180,37 @@
...
+    "sounds": [
+      {
+        "name": "galacticraft:machine/machinehum",
+        "stream": false
+      }
+    ]
+  },
+      "machine.oxygenfan": {
```
Roelymole:
```suggestion
  "machine.oxygenfan": {
```

src/main/java/dev/galacticraft/mod/client/sounds/GCSoundMap.java
TakeABao23: Is there a way this could be expressed as a switch statement, so it can be cleaned up in a similar way that the status types were?

Alternatively it could be a check for machine type like was briefly considered earlier.

src/main/java/dev/galacticraft/mod/client/sounds/GCSoundManager.java
```diff
@@ -0,0 +1,122 @@
...
+    // Finds a SoundInstance from a SoundEvent, if it exists and is currently playing
+    public Optional<MachineSound> getPlayingSoundInstance(SoundEvent soundEvent) {
+        for (var activeSound : this.activeSounds) {
+            // SoundInstances use their SoundEvent's id by default
+            if (activeSound.getLocation().equals(soundEvent.getLocation())) {
+                return Optional.of(activeSound);
+            }
+        }
+        return Optional.empty();
+    }
```
Roelymole: This method doesn't seem to be used anywhere. Also, it is likely that multiple elements in the `activeSound` list would share the same resource location, but this only matches the first one.

src/main/java/dev/galacticraft/mod/client/sounds/GCSoundMap.java
```diff
@@ -0,0 +1,55 @@
...
+        IDLE.put(GCMachineStatuses.NOT_GENERATING, null);
+        // Active sounds
+        ACTIVE.put(GCMachineStatuses.COLLECTING, GCSounds.MACHINE_HUM);
+        ACTIVE.put(GCMachineStatuses.DISTRIBUTING, GCSounds.MACHINE_HUM);
+        // All sounds
+        INTERNAL.put(false, IDLE);
+        INTERNAL.put(true, ACTIVE);
+        GC_SOUND_MAP = Collections.unmodifiableMap(INTERNAL);
+    }
+}
```
Roelymole, line 61: I've been thinking about this a bit more and I have a couple of thoughts. Making the maps private and adding a `get` method is probably a neater way of working.

Since `isActive` is an attribute of the machine status types there doesn't seem to be a need for the nested `IDLE` and `ACTIVE` maps.

We could then replace `GCSoundMap.GC_SOUND_MAP.get(isActive).getOrDefault(status, GCSounds.MACHINE_BUZZ)` with simply `GCSoundMap.get(status)`.

```java
public class GCSoundMap {
    private static final Map<MachineStatus, SoundEvent> GC_SOUND_MAP = Map.ofEntries(
            Map.entry(GCMachineStatuses.BLOCKED, null),
            Map.entry(GCMachineStatuses.NO_FUEL, null),
            Map.entry(GCMachineStatuses.NOT_ENOUGH_ENERGY, null),
            Map.entry(GCMachineStatuses.NOT_GENERATING, null),
            Map.entry(GCMachineStatuses.COLLECTING, GCSounds.MACHINE_HUM),
            Map.entry(GCMachineStatuses.DISTRIBUTING, GCSounds.MACHINE_HUM)
    );

    public static @Nullable SoundEvent get(MachineStatus status) {
        return GC_SOUND_MAP.getOrDefault(status, GCSounds.MACHINE_BUZZ);
    }
}
```

Alternatively, specifying the default values per type might be more flexible.

```java
public class GCSoundMap {
    private static final Map<MachineStatus.Type, SoundEvent> DEFAULTS = Map.ofEntries(
            Map.entry(MachineStatus.Type.WORKING, GCSounds.MACHINE_BUZZ),
            Map.entry(MachineStatus.Type.PARTIALLY_WORKING, GCSounds.MACHINE_BUZZ),
            Map.entry(MachineStatus.Type.MISSING_RESOURCE, GCSounds.MACHINE_BUZZ),
            Map.entry(MachineStatus.Type.MISSING_FLUIDS, GCSounds.MACHINE_BUZZ),
            Map.entry(MachineStatus.Type.MISSING_ENERGY, null),
            Map.entry(MachineStatus.Type.MISSING_ITEMS, GCSounds.MACHINE_BUZZ),
            Map.entry(MachineStatus.Type.OUTPUT_FULL, GCSounds.MACHINE_BUZZ),
            Map.entry(MachineStatus.Type.OTHER, GCSounds.MACHINE_BUZZ)
    );

    private static final Map<MachineStatus, SoundEvent> EXCEPTIONS = Map.ofEntries(
            Map.entry(GCMachineStatuses.BLOCKED, null),
            Map.entry(GCMachineStatuses.NO_FUEL, null),
            Map.entry(GCMachineStatuses.NOT_ENOUGH_ENERGY, null),
            Map.entry(GCMachineStatuses.NOT_GENERATING, null),
            Map.entry(GCMachineStatuses.COLLECTING, GCSounds.MACHINE_HUM),
            Map.entry(GCMachineStatuses.DISTRIBUTING, GCSounds.MACHINE_HUM)
    );

    public static @Nullable SoundEvent get(MachineStatus status) {
        return EXCEPTIONS.getOrDefault(status, DEFAULTS.get(status.getType()));
    }
}
```

TakeABao23: I like the second option! It has the best opportunity for scalability I think.

TakeABao23:
Ah it appears it won't build because it's using the MachineLib version that hasn't been committed yet

TakeABao23:
Is there anything left with this before it could be merged?

src/main/java/dev/galacticraft/mod/client/sounds/GCSoundMap.java
```diff
@@ -0,0 +1,57 @@
...
+ * SOFTWARE.
+ */
+
+package dev.galacticraft.mod.client.sounds;
+
+import dev.galacticraft.machinelib.api.machine.MachineStatus;
+import dev.galacticraft.machinelib.api.machine.MachineStatuses;
+import dev.galacticraft.mod.content.GCSounds;
+import dev.galacticraft.mod.machine.GCMachineStatuses;
+import me.shedaniel.cloth.clothconfig.shadowed.blue.endless.jankson.annotation.Nullable;
```
Roelymole: You are importing `Nullable` from the wrong place.

```suggestion
import org.jetbrains.annotations.Nullable;
```

src/main/java/dev/galacticraft/mod/client/sounds/GCSoundMap.java
```diff
@@ -0,0 +1,63 @@
...
+            Map.entry(GCMachineStatuses.NOT_GENERATING, Optional.empty()),
+            Map.entry(GCMachineStatuses.COLLECTING, Optional.of(GCSounds.MACHINE_HUM)),
+            Map.entry(GCMachineStatuses.DISTRIBUTING, Optional.of(GCSounds.MACHINE_HUM)));
+
+    private static final Map<MachineStatus.Type, Optional<SoundEvent>> COALEXCEPTIONS = Map.ofEntries();
+
+    public static @Nullable SoundEvent get(MachineStatus status, MachineBlockEntity machine) {
+        if (machine.getClass() == CompressorBlockEntity.class || machine.getClass() == CoalGeneratorBlockEntity.class) {
+            return COALEXCEPTIONS.get(status.getType()).orElse(null);
+        }
```
Roelymole: I would prefer the name `COAL_EXCEPTIONS`, although it might be better to remove this and just return null until we find ourselves needing it.

It might be neater to add an interface (it doesn't matter if it is empty) such as `CoalPoweredMachine` that `CompressorBlockEntity` and `CoalGeneratorBlockEntity` both implement, then you could do use `if (machine instanceof CoalPoweredMachine) {...}`

src/main/java/dev/galacticraft/mod/client/sounds/GCSoundMap.java
```diff
@@ -0,0 +1,63 @@
...
+import dev.galacticraft.machinelib.api.machine.MachineStatus;
+import dev.galacticraft.machinelib.api.machine.MachineStatuses;
+import dev.galacticraft.mod.content.GCSounds;
+import dev.galacticraft.mod.content.block.entity.machine.CoalGeneratorBlockEntity;
+import dev.galacticraft.mod.content.block.entity.machine.CompressorBlockEntity;
+import dev.galacticraft.mod.machine.GCMachineStatuses;
+import java.util.Map;
+import java.util.Optional;
+import org.jetbrains.annotations.Nullable;
+import net.minecraft.sounds.SoundEvent;
```
Roelymole: These imports need rearranging again, thanks.

```suggestion
import net.minecraft.sounds.SoundEvent;
import org.jetbrains.annotations.Nullable;

import java.util.Map;
import java.util.Optional;
```

src/main/java/dev/galacticraft/mod/client/sounds/GCSoundManager.java
```diff
@@ -0,0 +1,96 @@
...
+    public static void onStatusChanged(Minecraft minecraft, LocalPlayer player, BlockPos pos, MachineStatus status, MachineStatus oldStatus) {
+        MachineBlockEntity machine = (MachineBlockEntity) minecraft.level.getBlockEntity(pos);
+        GCSoundManager manager = GCSoundManager.getInstance();
+        boolean isActive = status.getType().isActive();
+        float maxVolume = isActive ? 1.0F : 0.2F;
+        // Stop old sound (if there is one)
+        manager.getSoundFromEntity(machine, oldStatus, isActive).ifPresent(oldSound -> oldSound.end());
+        // Play new sound (if there is one)
+        SoundEvent newSound = GCSoundMap.get(status, machine);
+        manager.play(new MachineSound(machine, newSound, manager, maxVolume));
```
Roelymole, line 95: It might be best to only call the play method if `newSound` is not null, although it doesn't seem to cause any errors as it currently stands.

TakeABao23: good point

gradle.properties
```diff
@@ -19,7 +19,7 @@ rei.version=16.0.754
 jei.version=19.21.0.247
 emi.version=1.1.22+1.21.1
 dynamicdimensions.version=0.9.1+34
-machinelib.version=0.7.0+142
+machinelib.version=0.7.0+62641606
```
Roelymole: The version you need is available on the Maven now.

```suggestion
machinelib.version=0.7.0+144
```

src/main/java/dev/galacticraft/mod/client/sounds/GCSoundMap.java
```diff
@@ -0,0 +1,61 @@
...
+            Map.entry(GCMachineStatuses.BLOCKED, Optional.empty()),
+            Map.entry(GCMachineStatuses.NOT_GENERATING, Optional.empty()),
+            Map.entry(GCMachineStatuses.COLLECTING, Optional.of(GCSounds.MACHINE_HUM)),
+            Map.entry(GCMachineStatuses.DISTRIBUTING, Optional.of(GCSounds.MACHINE_HUM)));
+
+    public static @Nullable SoundEvent get(MachineStatus status, MachineBlockEntity machine) {
+        if (machine instanceof CoalPoweredMachine) {
+            return null;
+        }
+        return EXCEPTIONS.getOrDefault(status, DEFAULTS.get(status.getType())).orElse(null);
```
Roelymole: What I told you isn't the most future-proof thing to do here.

It would be safer if it was `DEFAULTS.getOrDefault(status.getType(), Optional.empty())` to avoid a null pointer exception if the type isn't present in the map (which could happen if a new type is introduced).

Roelymole (approved):
Provided the workflow passes (using the correct MachineLib version this time), then I'm happy to merge this in.

PR #573, opened 2026-04-29
Feature: Fluid Canister

HarryPotato986:
The Fluid Canister is component driven. It updates it's name, tooltip, window colour, and fill level based on the type and amount of fluid it stores.

For fluids that use tints, it can automatically get their colour. For Fluids with the colour baked into the texture (Like galacticraft fuel and oil), custom colours can be added to the ColorProvider in GalacticraftClient.

I've given the legacy fluids their legacy colours, but these could be changed later as mentioned.
I've also added the empty, crude oil, methane, and liquid oxygen canisters to the proper creative mode tab. The code is there for liquid nitrogen, but commented out as the fluid does not yet exist.
I've also added the crafting recipe and a fluid tag for fluids that should be incompatible with the fluid canister.


src/main/java/dev/galacticraft/mod/data/model/GCModelProvider.java
```diff
@@ -953,6 +955,16 @@ public final void createOlivineCluster(BlockModelGenerators generator, Block blo
                 );
     }
 
+    public void generateFluidCanisterModels(ItemModelGenerators generator) {
+        for (int i = 1; i <= 6; i++) {
+            generator.generateLayeredItem(
+                    ModelLocationUtils.getModelLocation(GCItems.FLUID_CANISTER).withSuffix("_"+i),
+                    Constant.id("item/fluid_canister_base"),
+                    Constant.id("item/fluid_canister_overlay_"+i)
+            );
```
Roelymole, line 965: You should be able to use the same `ResourceLocation` with different suffixes.

```suggestion
            
            ResourceLocation resourceLocation = ModelLocationUtils.getModelLocation(GCItems.FLUID_CANISTER);
            generator.generateLayeredItem(
                    resourceLocation.withSuffix("_" + i),
                    resourceLocation.withSuffix("_base"),
                    resourceLocation.withSuffix("_overlay_" + i)
            );
```

src/main/java/dev/galacticraft/mod/Constant.java
```diff
@@ -520,6 +520,8 @@ interface Item {
         String BATTERY = "battery";
         String INFINITE_BATTERY = "infinite_battery";
         String INFINITE_INDICATOR = "infinite_indicator";
+        String FLUID_CANISTER = "fluid_canister";
+        String FlUID_CANISTER_WINDOW = "fluid_canister_window";
```
Roelymole: You have used a lower case L in the name of this constant. As far as I can see it isn't being used anywhere either.

src/main/java/dev/galacticraft/mod/content/GCFluidStorageProviders.java
```diff
@@ -0,0 +1,33 @@
...
+
+package dev.galacticraft.mod.content;
+
+import dev.galacticraft.mod.content.item.GCItems;
+import dev.galacticraft.mod.storage.CanisterFluidStorage;
+import net.fabricmc.fabric.api.transfer.v1.fluid.FluidStorage;
+
+public class GCFluidStorageProviders {
+    public static void register() {
+        FluidStorage.ITEM.registerForItems((itemStack, context) -> new CanisterFluidStorage(context), GCItems.FLUID_CANISTER);
```
Roelymole: This could probably go in the `GCApiLookupProviders` class.

src/main/java/dev/galacticraft/mod/content/item/FluidCanisterItem.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    @Override
+    public void appendHoverText(ItemStack stack, TooltipContext context, List<Component> tooltip, TooltipFlag flag) {
+        FluidData data = stack.get(FLUID_DATA);
+
+        if (data != null && !data.variant().isBlank()) {
+            Component fluidName;
+            if (data.variant().getFluid().isSame(GCFluids.LIQUID_OXYGEN)) {
+                fluidName = Component.literal("LOX");
```
Roelymole: This abbreviation should still probably be translatable.

src/main/java/dev/galacticraft/mod/content/item/FluidCanisterItem.java
Roelymole, line 1: You have defined `FLUID_CANISTER_FILLED`, `FLUID_CANISTER_EMPTY` and `FLUID_CANISTER_FLUID_INFO` in the `Translations` interface, but you aren't using them.

src/main/java/dev/galacticraft/mod/content/item/FluidCanisterItem.java
```diff
@@ -0,0 +1,105 @@
...
+        FluidData data = stack.get(FLUID_DATA);
+
+        if (data == null || data.variant().isBlank()) {
+            return Component.translatable("item.galacticraft.fluid_canister");
+        }
+
+        Component fluidName = FluidVariantAttributes.getName(data.variant()).plainCopy();
+
+        return Component.translatable("item.galacticraft.fluid_canister.filled", fluidName);
+    }
```
Roelymole, line 77: I would suggest reordering this so that the superclass method is used as a fallback if the fluid variant cannot be found.

```java
    @Override
    public @NotNull Component getName(ItemStack stack) {
        FluidData data = stack.get(FLUID_DATA);

        if (data != null && !data.variant().isBlank()) {
            Component fluidName = FluidVariantAttributes.getName(data.variant()).plainCopy();
            return Component.translatable(Translations.FLUID_CANISTER_FILLED, fluidName);
        }

        return super.getName(stack);
    }
```

src/main/java/dev/galacticraft/mod/content/item/GCCreativeModeTabs.java
```diff
@@ -293,6 +298,24 @@ public class GCCreativeModeTabs {
...
+                output.accept(fuelCanister);
+                ItemStack methaneCanister = new ItemStack(FLUID_CANISTER);
+                methaneCanister.set(FLUID_DATA, new FluidData(FluidVariant.of(Gases.METHANE), 81000));
+                output.accept(methaneCanister);
+                ItemStack liquidOxygenCanister = new ItemStack(FLUID_CANISTER);
+                liquidOxygenCanister.set(FLUID_DATA, new FluidData(FluidVariant.of(GCFluids.LIQUID_OXYGEN), 81000));
+                output.accept(liquidOxygenCanister);
+                //ItemStack liquidNitrogenCanister = new ItemStack(FLUID_CANISTER);
+                //liquidNitrogenCanister.set(FLUID_DATA, new FluidData(FluidVariant.of(GCFluids.LIQUID_NITROGEN), 81000)); // Liquid nitrogen not added yet
+                //output.accept(liquidNitrogenCanister);
```
Roelymole: It would be more maintainable to define the maximum capacity of fluid canisters in a single place and reference that value elsewhere.

Also, since this code is repeated several times, it would be beneficial to move it to a static method in `FluidCanisterItem` similarly to the `OxygenTankItem.getFullTank`.

src/main/java/dev/galacticraft/mod/GalacticraftClient.java
```diff
@@ -230,6 +237,37 @@ public void onInitializeClient() {
...
+            int color = FluidVariantRendering.getColor(data.variant());
+
+            if (color == -1) {
+                if (data.variant().isOf(Fluids.WATER)) return 0xFF3F76E4;
+                if (data.variant().isOf(GCFluids.CRUDE_OIL)) return 0xFF171313;
+                if (data.variant().isOf(GCFluids.FUEL)) return 0xFFB8D200;
+                if (data.variant().isOf(Gases.METHANE)) return 0xFF80FFFF;
+                if (data.variant().isOf(GCFluids.LIQUID_OXYGEN)) return 0xFFD76453;
+                //if (data.variant().isOf(GCFluids.LIQUID_NITROGEN)) return 0xFF002BD2; // Not added yet
+                if (data.variant().isOf(GCFluids.SULFURIC_ACID)) return 0xFF99FF33;
```
Roelymole: It would be better to set the tint color for our fluids in the `GalacticraftClient` class.

src/main/resources/assets/galacticraft/textures/item/fluid_canister_base.png
Roelymole, line 1: The existing `empty_liquid_canister.png`, `liquid_canister.png` and `oil_canister_partial_<1-6>.png` textures are redundant, so should be deleted.

src/main/java/dev/galacticraft/mod/content/item/FluidCanisterItem.java
```diff
@@ -0,0 +1,105 @@
...
+
+    @Override
+    public void appendHoverText(ItemStack stack, TooltipContext context, List<Component> tooltip, TooltipFlag flag) {
```
Roelymole, line 80: Please could you do something more similar to [OxygenTankItem](https://github.com/TeamGalacticraft/Galacticraft/blob/main/src/main/java/dev/galacticraft/mod/content/item/OxygenTankItem.java#L107-L116) for consistency.

src/main/java/dev/galacticraft/mod/data/recipes/GCGearRecipeProvider.java
```diff
@@ -323,6 +324,20 @@ public void buildRecipes(RecipeOutput output) {
                 .unlockedBy(getHasName(GCItems.COMPRESSED_STEEL), has(GCItems.COMPRESSED_STEEL))
                 .emiDefaultRecipe(true)
                 .save(output);
+
+        GCShapedRecipeBuilder.crafting(RecipeCategory.TOOLS, GCItems.FLUID_CANISTER)
+                .define('T', GCItems.COMPRESSED_TIN)
+                .define('S', GCItems.COMPRESSED_STEEL)
+                .define('G', Items.GLASS)
+                .define('C', GCItems.TIN_CANISTER)
```
Roelymole: It would be better to use `ConventionalItemTags.GLASS_BLOCKS_COLORLESS` here instead and maybe `GCItemTags.TIN_CANISTERS` too.

src/main/java/dev/galacticraft/mod/storage/CanisterFluidStorage.java
```diff
@@ -0,0 +1,87 @@
...
+import net.minecraft.world.item.ItemStack;
+
+import static dev.galacticraft.api.component.GCDataComponents.FLUID_DATA;
+
+public class CanisterFluidStorage extends SingleVariantItemStorage<FluidVariant> {
+    private final ContainerItemContext itemContext;
+
+    public CanisterFluidStorage(ContainerItemContext context) {
+        super(context);
+        this.itemContext = context;
```
Roelymole: Rather than storing the whole context, it would be better to store what you are actually using.

```suggestion
        this.fluidData = context.getItemVariant().toStack().get(FLUID_DATA);
```

src/main/java/dev/galacticraft/mod/content/item/FluidCanisterItem.java
```diff
@@ -0,0 +1,105 @@
...
+
+    @Override
+    public boolean isBarVisible(ItemStack stack) {
+        return true;
+    }
+
+    @Override
+    public int getBarWidth(ItemStack stack) {
+        StorageView<FluidVariant> storage = FluidCanisterItem.getStorage(stack);
+        return Math.round(13.0F - (float) (storage.getCapacity() - storage.getAmount()) * 13.0F / (float) storage.getCapacity());
```
Roelymole: This can be simplified (like 311b62d4bf228eff1138caa8736c6902e351a5db).

```suggestion
        return Math.round(13.0F * ((float) storage.getAmount() / (float) storage.getCapacity()));
```

src/main/java/dev/galacticraft/mod/tag/GCFluidTags.java
```diff
@@ -38,6 +38,7 @@ public class GCFluidTags {
     public static final TagKey<Fluid> LIQUID_OXYGEN = commonTag("oxygen");
     public static final TagKey<Fluid> OXYGEN = galacticraftTag("oxygen");
     public static final TagKey<Fluid> NON_BREATHABLE = galacticraftTag("non_breathable");
+    public static final TagKey<Fluid> FLUID_CANISTER_EXCLUDED = galacticraftTag("fluid_canister_excluded");
```
Roelymole: I think the word exclusions might read a bit better than excluded.

src/main/java/dev/galacticraft/mod/GalacticraftClient.java
Roelymole: Rather than hardcoding these values (which won't work for fluids from other mods without a tint color) using something like `TextureUtils.calculateAverageColor` might be better (it would need to be made public first).

```suggestion
                TextureAtlasSprite sprite = FluidVariantRendering.getSprite(data.variant());
                if (sprite != null) {
                    color = FastColor.ARGB32.opaque(TextureUtils.calculateAverageColor(sprite));
                }
```

It might be good to cache these colors and register a callback to the `InvalidateRenderStateCallback.EVENT` to invalidate the cache.

Roelymole (changes requested):
This is looking good, but there are a few things that could do with changing first.

src/main/java/dev/galacticraft/mod/content/item/FluidCanisterItem.java
HarryPotato986: These are being used in GCTranslationProvider.

src/main/java/dev/galacticraft/mod/GalacticraftClient.java
HarryPotato986: Will methane and liquid oxygen be getting textures or should I still hardcode their colours?

In my opinion it would be nice to keep all five legacy colours hardcoded.
```java
ColorProviderRegistry.ITEM.register((stack, layer) -> {
    if (layer != 1) return -1;

    FluidData data = stack.get(FLUID_DATA);
    if (data == null || data.variant().isBlank()) return -1;

    return FLUID_CANISTER_COLOR_CACHE.computeIfAbsent(data.variant(), variant -> {
        // Try the tint
       int color = FluidVariantRendering.getColor(data.variant());
       if (color != -1) return color;

       // Manual colors
       if (data.variant().isOf(GCFluids.CRUDE_OIL)) return 0xFF171313;
       if (data.variant().isOf(GCFluids.FUEL)) return 0xFFB8D200;
       if (data.variant().isOf(Gases.METHANE)) return 0xFF80FFFF;
       if (data.variant().isOf(GCFluids.LIQUID_OXYGEN)) return 0xFFD76453;
       //if (data.variant().isOf(GCFluids.LIQUID_NITROGEN)) return 0xFF002BD2; // Not added yet

       // Try to average the texture color
       TextureAtlasSprite sprite = FluidVariantRendering.getSprite(variant);
       if (sprite != null) {
           return FastColor.ARGB32.opaque(TextureUtils.calculateAverageColor(sprite));
       }
       return -1;
    });
}, GCItems.FLUID_CANISTER);
InvalidateRenderStateCallback.EVENT.register(FLUID_CANISTER_COLOR_CACHE::clear);
```

src/main/java/dev/galacticraft/mod/GalacticraftClient.java
```diff
@@ -230,6 +244,44 @@ public void onInitializeClient() {
...
+                // Try the tint
+                int color = FluidVariantRendering.getColor(data.variant());
+                if (color != -1) return color;
+
+                // Manual colors
+                if (data.variant().isOf(GCFluids.CRUDE_OIL)) return 0xFF171313;
+                if (data.variant().isOf(GCFluids.FUEL)) return 0xFFB8D200;
+                if (data.variant().isOf(Gases.METHANE)) return 0xFF80FFFF;
+                if (data.variant().isOf(GCFluids.LIQUID_OXYGEN)) return 0xFFD76453;
+                //if (data.variant().isOf(GCFluids.LIQUID_NITROGEN)) return 0xFF002BD2; // Not added yet
```
Roelymole: I was thinking these would be removed since they don't match the current textures anyway.

src/main/java/dev/galacticraft/mod/content/item/FluidCanisterItem.java
Roelymole: I meant you were still hardcoding the translation keys rather than using the constants, which you have fixed now.

src/main/java/dev/galacticraft/mod/GalacticraftClient.java
Roelymole: I don't think there is a problem hardcoding the colours for fluids that don't have textures yet, but for fluids with textures I would prefer not to use the old colours since they don't match anymore.

src/main/java/dev/galacticraft/mod/content/item/FluidCanisterItem.java
Roelymole: Maybe I didn't make myself quite clear. I was meaning to use the `TooltipUtil.appendFluidRemainingTooltip` method, or copying the relevant parts if that doesn't quite do what you are after.

In particular, `TooltipUtil.formatFluidRemaining` handles the units to display the amount/capacity in (showing the raw values when shift is held down).

It also formats the numbers with a comma every 3 digits, so 1000 becomes 1,000 which is consistent with how the fluid tank tooltips are displayed in the GUIs of our machines.

src/main/java/dev/galacticraft/mod/content/item/GCItems.java
```diff
@@ -175,6 +175,8 @@ public class GCItems {
 
     public static final Item STANDARD_WRENCH = ITEMS.register(Constant.Item.STANDARD_WRENCH, new StandardWrenchItem(new Item.Properties().durability(256)));
 
+    public static final Item FLUID_CANISTER = ITEMS.register(Constant.Item.FLUID_CANISTER, new FluidCanisterItem(new Item.Properties().stacksTo(1)));
```
Roelymole: In the spirit of making things more data-driven, I think it is worth adding the fluid canister capacity to the config and then passing in the value of `Galacticraft.CONFIG.fluidCanisterCapacity()` to a `capacity` argument.

This is what is being done for the oxygen tank items if you want an example.

src/main/java/dev/galacticraft/mod/data/recipes/GCGearRecipeProvider.java
```diff
@@ -323,6 +323,20 @@ public void buildRecipes(RecipeOutput output) {
...
+                .define('T', GCItems.COMPRESSED_TIN)
+                .define('S', GCItems.COMPRESSED_STEEL)
+                .define('G', ConventionalItemTags.GLASS_BLOCKS_COLORLESS)
+                .define('C', GCItemTags.TIN_CANISTERS)
+                .pattern("TST")
+                .pattern("TGT")
+                .pattern("TCT")
+                .unlockedBy(getHasName(GCItems.COMPRESSED_TIN), has(GCItems.COMPRESSED_TIN))
+                .unlockedBy(getHasName(GCItems.COMPRESSED_STEEL), has(GCItems.COMPRESSED_STEEL))
+                .unlockedBy(getHasName(GCItems.TIN_CANISTER), has(GCItems.TIN_CANISTER))
```
Roelymole: It should probably be `has(GCItemTags.TIN_CANISTERS)` here.

src/main/java/dev/galacticraft/mod/storage/CanisterFluidStorage.java
```diff
@@ -0,0 +1,85 @@
...
+import net.minecraft.world.item.ItemStack;
+
+import static dev.galacticraft.api.component.GCDataComponents.FLUID_DATA;
+
+public class CanisterFluidStorage extends SingleVariantItemStorage<FluidVariant> {
+    private final FluidData fluidData;
+
+    public CanisterFluidStorage(ContainerItemContext context) {
+        super(context);
+        this.fluidData = context.getItemVariant().toStack().get(FLUID_DATA);;
```
Roelymole: There is a superfluous semi-colon at the end of this line.

src/main/java/dev/galacticraft/mod/data/GCTranslationProvider.java
```diff
@@ -975,6 +978,10 @@ protected void generateTooltipTranslations() {
         this.add(Tooltip.TIME_UNTIL_COOL, "Time Until Cool: %s");
         this.add(Tooltip.SECONDS_UNIT, "%ss");
         this.add(Tooltip.INCORRECT_NUMBER_OF_SLOTS, "Incorrect number of slots in the workstation for this recipe.");
+
+        this.add(Tooltip.FLUID_CANISTER_EMPTY, "Empty");
+        this.add(Tooltip.FLUID_CANISTER_FLUID_INFO, "%s: %d");
```
Roelymole: This should be `"%s: %s"` now that you are substituting in a string, not an integer.

src/main/java/dev/galacticraft/mod/config/ConfigImpl.java
```diff
@@ -778,6 +788,18 @@ public Screen create(Screen parent) {
                     .build()
             );
 
+            client.addEntry(new LongFieldBuilder(
```
Roelymole: This isn't just a client-side option, if nothing else it could go in the misc category.

src/main/java/dev/galacticraft/mod/storage/CanisterFluidStorage.java
```diff
@@ -35,10 +35,12 @@
...
 public class CanisterFluidStorage extends SingleVariantItemStorage<FluidVariant> {
     private final FluidData fluidData;
+    private final long capacity;
 
-    public CanisterFluidStorage(ContainerItemContext context) {
+    public CanisterFluidStorage(ItemStack stack, ContainerItemContext context) {
         super(context);
-        this.fluidData = context.getItemVariant().toStack().get(FLUID_DATA);;
+        this.fluidData = context.getItemVariant().toStack().get(FLUID_DATA);
+        this.capacity = ((FluidCanisterItem) stack.getItem()).capacity;
```
Roelymole: Is the `stack` argument being provided not the same as `context.getItemVariant().toStack()`?

If they are representing the same thing then it would be better if they both used the same approach.

src/main/java/dev/galacticraft/mod/util/Translations.java
```diff
@@ -153,6 +153,7 @@ interface Config {
 
         String CLIENT = "config.galacticraft.client";
         String SQUARE_CANNED_FOOD = "config.galacticraft.client.square_canned_food";
+        String FLUID_CANISTER_CAPACITY = "config.galacticraft.client.fluid_canister_capacity";
```
Roelymole: Again, not really a client-side option.

src/main/java/dev/galacticraft/mod/content/item/FluidCanisterItem.java
```diff
@@ -0,0 +1,107 @@
...
+    @Override
+    public void appendHoverText(ItemStack stack, TooltipContext context, List<Component> tooltip, TooltipFlag flag) {
+        this.appendFluidCanisterTooltip(stack, context, tooltip, flag);
+        super.appendHoverText(stack, context, tooltip, flag);
+    }
+
+    protected void appendFluidCanisterTooltip(ItemStack stack, TooltipContext context, List<Component> tooltip, TooltipFlag type) {
+        FluidData data = stack.get(FLUID_DATA);
+        TooltipUtil.appendCanisterRemainingTooltip(Tooltip.FLUID_CANISTER_FLUID_INFO, data, capacity, tooltip);
+    }
```
Roelymole: The main reason `OxygenTankItem` has the separate `appendOxygenTankTooltip` method is so that it can be overriden in the `InfiniteOxygenTank` class, which isn't necessary here, so the body of this method could be moved back into `appendHoverText`.

src/main/java/dev/galacticraft/mod/util/TooltipUtil.java
```diff
@@ -106,6 +114,20 @@ public static void appendFluidRemainingTooltip(String resourceId, long amount, l
...
+        if (data != null && !data.variant().isBlank()) {
+            Component fluidName;
+            if (data.variant().getFluid().isSame(GCFluids.LIQUID_OXYGEN)) {
+                fluidName = Component.translatable(Translations.Tooltip.FLUID_CANISTER_LOX);
+            } else {
+                fluidName = FluidVariantAttributes.getName(data.variant()).plainCopy();
+            }
+            appendLabeledTooltip(resourceId, fluidName, TooltipUtil.formatFluidRemaining(data.amount(), capacity), tooltip);
+        } else {
+            tooltip.add(Component.translatable(Translations.Tooltip.FLUID_CANISTER_EMPTY).withStyle(ChatFormatting.DARK_GRAY));
```
Roelymole: Maybe use `setStyle(Constant.Text.DARK_GRAY_STYLE)` instead of `withStyle(ChatFormatting.DARK_GRAY)` to be consistent with the rest of the class.

Roelymole (commented):
There are a couple last things that it would be good to change and then I'm happy to merge this in.

Roelymole (approved):

PR #577, opened 2026-05-28
Fix/server stability

maxryan008:


src/main/java/dev/galacticraft/mod/content/block/decoration/CannedFoodBlock.java
```diff
@@ -131,38 +126,14 @@ protected BlockState updateShape(BlockState blockState, Direction direction, Blo
...
-                    AABB shape = new AABB(x, y, z, x + 0.375D, y + 0.5D, z + 0.375D);
-                    double dist = shape.distanceToSqr(location);
-                    if (dist < minDist) {
-                        minDist = dist;
-                        itemStack = canContents.get(i);
-                    }
-                }
-
-                if (CannedFoodItem.getSize(itemStack) > 0) {
-                    return itemStack;
```
Roelymole, line 162: It is a shame that this doesn't work on a server, but could you move this code to a WAILA/WTHIT plugin so that it still displays the correct tooltip at least.

src/main/java/dev/galacticraft/mod/content/item/CannedFoodLayout.java
Roelymole, line 1: Did you consider moving this to `CannedFoodBlock`, since that should be available on the client and the server?

Especially since it relates more to the block form than the item form.

src/main/java/dev/galacticraft/mod/content/item/CannedFoodItem.java
Roelymole, line 1: Maybe the data component for the can colour should only be used as an override (I do quite like the random pastel colours though).

We should be able to do something similar to the fluid canister tints with caching ([see here](https://github.com/TeamGalacticraft/Galacticraft/blob/5a4050a8be1d04cb9d716dbeecf0e86b4d2f281d/src/main/java/dev/galacticraft/mod/GalacticraftClient.java#L251-L276)).

I think the most fiddly part will be working out the best way to convert a list of item stacks into a key.

src/main/java/dev/galacticraft/mod/network/c2s/CapeSelectionPayload.java
```diff
@@ -57,17 +59,35 @@ public record CapeSelectionPayload(CapeMode mode, String gcCapeId) implements C2
...
-            id = null;
+        CapeMode rawMode = this.mode == null ? CapeMode.VANILLA : this.mode;
+        String rawId = this.gcCapeId;
+
+        String playerName = player.getGameProfile().getName();
+        String uuid = player.getUUID().toString();
+        CapeRole role = CapesLoader.roleFor(uuid);
+
+        CapeMode finalMode = rawMode;
+        String finalId = rawId;
```
Roelymole: Having separate `rawMode`/`finalMode` and `rawId`/`finalId` variables isn't currently necessary.

src/main/java/dev/galacticraft/mod/network/c2s/CapeSelectionPayload.java
```diff
@@ -77,7 +97,27 @@ public void handle(ServerPlayNetworking.@NotNull Context context) {
             ));
         }
 
-        context.server().getPlayerList().broadcastAll(ServerPlayNetworking.createS2CPacket(new CapeAssignmentsPacket(list)));
+        ServerPlayNetworking.send(target, new CapeAssignmentsPacket(list));
+    }
+
+    public static void broadcastCapeSnapshot(net.minecraft.server.MinecraftServer server) {
```
Roelymole: You are already importing `net.minecraft.server.MinecraftServer`, so this can be simplified to just `MinecraftServer`.

src/main/java/dev/galacticraft/mod/network/c2s/CapeSelectionPayload.java
```diff
@@ -57,17 +59,35 @@ public record CapeSelectionPayload(CapeMode mode, String gcCapeId) implements C2
...
+            finalMode = CapeMode.VANILLA;
+            finalId = null;
+        }
+
+        ServerCapeManager.set(player, finalMode, finalId);
+
+        broadcastCapeSnapshot(context.server());
+    }
+
+    public static void sendCapeSnapshot(net.minecraft.server.level.ServerPlayer target) {
```
Roelymole: You are already importing `net.minecraft.server.level.ServerPlayer`, so this can be simplified to just `ServerPlayer`.

src/main/java/dev/galacticraft/mod/network/s2c/CapeAssignmentsPacket.java
```diff
@@ -81,9 +81,14 @@ public Entry(String uuid, CapeMode mode, String gcCapeId) {
     public Runnable handle(@NotNull ClientPlayNetworking.Context context) {
         return () -> {
             var map = new HashMap<String, CapesClientState.Entry>(entries.size());
+
             for (var e : entries) {
-                map.put(e.uuid.toLowerCase(), new CapesClientState.Entry(e.mode, e.gcCapeId));
+                map.put(
+                        e.uuid.toLowerCase(java.util.Locale.ROOT),
```
Roelymole, line 87: Please could you import `java.util.Locale`

src/main/java/dev/galacticraft/mod/network/c2s/SatelliteCreationPayload.java
```diff
@@ -50,21 +50,66 @@ public record SatelliteCreationPayload(ResourceKey<CelestialBody<?, ?>> body) im
 
     @Override
     public void handle(ServerPlayNetworking.@NotNull Context context) {
+        if (!dev.galacticraft.mod.Galacticraft.CONFIG.enableSpaceStationCreation()) {
```
Roelymole: Please could you import `dev.galacticraft.mod.Galacticraft`

src/main/java/dev/galacticraft/mod/client/gui/screen/ingame/CelestialSelectionScreen.java
```diff
@@ -975,14 +1198,19 @@ protected void renderCelestialScreenTooltip(GuiGraphics gui, Graphics graphics,
...
             }
         }
-        return i;
+
+        return amount;
+    }
+
+    private static boolean isInside(double mouseX, double mouseY, double x, double y, double width, double height) {
+        return mouseX >= x && mouseX < x + width && mouseY >= y && mouseY < y + height;
     }
```
Roelymole: This is a duplicate of `DrawableUtil.mouseIn`

src/main/java/dev/galacticraft/mod/client/gui/screen/ingame/CelestialSelectionScreen.java
```diff
@@ -957,12 +1182,10 @@ protected void renderCelestialScreenTooltip(GuiGraphics gui, Graphics graphics,
         gui.pose().translate(0, 0, BORDER_Z + 1);
 
         try (Graphics.Fill fill = graphics.fill()) {
-            // Draw the tooltip base which is a rectangle missing the corners
             fill.fillGradientRaw(x1 + 1, y1, x2 - 1, y2, colorA, colorA);
             fill.fillGradientRaw(x1, y1 + 1, x1 + 1, y2 - 1, colorA, colorA);
             fill.fillGradientRaw(x2 - 1, y1 + 1, x2, y2 - 1, colorA, colorA);
 
-            // Inner border: left, right, top, bottom (lighter at the top, darker at the bottom)
```
Roelymole, line 965: I don't quite understand why you have deleted these comments.

It isn't immediately clear (to me at least) what this code corresponds to, so I thought they were quite helpful.

src/main/java/dev/galacticraft/mod/client/gui/screen/ingame/CelestialSelectionScreen.java
```diff
@@ -823,98 +994,136 @@ private void drawSpaceStationCreationPrompt(GuiGraphics gui, Graphics graphics,
...
             }
         }
+
         return list;
     }
 
-    /**
-     * Draws child bodies (when appropriate) on the left-hand interface
-     */
+    private java.util.Map<ResourceLocation, CelestialBody<SatelliteConfig, SatelliteType>> satellites() {
```
Roelymole: Please could you import `java.util.Map`

src/main/java/dev/galacticraft/mod/misc/cape/CapesClientRole.java
```diff
@@ -48,52 +50,70 @@ private static final class PlayerRole {
     public static void ensureLoadedAsync() {
         if (LOADED.get()) return;
         LOADED.set(true);
-        // load off-thread
+
         new Thread(() -> {
             try (var in = new URL(Constant.CAPES).openStream();
                  var r = new InputStreamReader(in, StandardCharsets.UTF_8)) {
```
Roelymole: This line is indented by one more space than it should be.

src/main/java/dev/galacticraft/mod/misc/cape/ServerCapeManager.java
```diff
@@ -42,21 +41,49 @@ public Assignment(CapeMode mode, String gcCapeId) {
...
     }
+
+    public static void remove(ServerPlayer player) {
+        if (player == null) return;
+        ACTIVE.remove(player.getUUID().toString().toLowerCase(Locale.ROOT));
+    }
+
+    public static boolean rolesLoaded() {
+        return CapesLoader.isLoaded();
+    }
```
Roelymole: As far as I can see this is not being used.

src/main/java/dev/galacticraft/mod/client/gui/screen/ingame/CelestialSelectionScreen.java
```diff
@@ -79,656 +89,817 @@ public class CelestialSelectionScreen extends CelestialScreen {
...
+            boolean canCreateStations,
+            CelestialBody<?, ?> fromBody,
+            Collection<ResourceLocation> disabledDestinations
+    ) {
         super(Component.empty());
         this.mapMode = mapMode;
         this.data = data;
         this.canCreateStations = canCreateStations;
         this.fromBody = fromBody;
+        this.disabledDestinations = new HashSet<>(disabledDestinations);
```
Roelymole: This line causes a crash when `disabledDestinations` is null, which is the case where it is being used in `GCKeyBinds` and maybe elsewhere.

Roelymole (commented):
I'm not quite sure what the issues you were having with the capes were, but the changes seem reasonable.

src/main/java/dev/galacticraft/mod/misc/cape/CapesClientRole.java
maxryan008: it isnt. the open bracket in try (var in =... does not get closed until the end of var r = new.... However i will make it a bit more obvious and seperate the open bracket with a new line

Roelymole: Oh, I see, that is much clearer now.

src/main/java/dev/galacticraft/mod/GalacticraftClient.java
```diff
@@ -65,7 +65,7 @@
 import dev.galacticraft.mod.screen.GCMenuTypes;
 import dev.galacticraft.mod.screen.GCPlayerInventoryMenu;
 import dev.galacticraft.mod.screen.RocketMenu;
-import dev.galacticraft.mod.util.TextureUtils;
+import dev.galacticraft.mod.client.util.TextureUtils;
```
Roelymole: This import should probably be moved to be with the rest of the imports from the `dev.galacticraft.mod.client.util` package.

src/main/java/dev/galacticraft/mod/client/model/CannedFoodBakedModel.java
```diff
@@ -22,8 +22,8 @@
 
 package dev.galacticraft.mod.client.model;
 
-import dev.galacticraft.api.component.GCDataComponents;
 import dev.galacticraft.mod.content.block.decoration.CannedFoodBlock;
+import dev.galacticraft.mod.client.util.CannedFoodColorCache;
```
Roelymole, line 25: These two imports should probably be the other way round so that they are in alphabetical order.

src/main/java/dev/galacticraft/mod/client/util/CannedFoodColorCache.java
Roelymole, line 1: It might be worth considering if there are any changes in #439 that we would want to include here, but it is not a blocking issue either way.

src/main/java/dev/galacticraft/mod/misc/cape/CapesClientState.java
```diff
@@ -48,8 +49,27 @@ public static synchronized void apply(Map<String, Entry> snapshot) {
...
+
+        if (entry != null) {
+            return entry;
+        }
+
+        var mc = Minecraft.getInstance();
+
+        if (mc.player != null
+                && mc.player.getGameProfile() != null
+                && mc.player.getGameProfile().getId().equals(player.getGameProfile().getId())) {
```
Roelymole: This can be simplified a bit, since we have already checked that `player.getGameProfile()` is not null and the equals method will return false if the argument is null.

```suggestion
        if (mc.player != null && player.getGameProfile().getId().equals(mc.player.getGameProfile())) {
```

src/main/java/dev/galacticraft/mod/client/util/CannedFoodColorCache.java
```diff
@@ -0,0 +1,207 @@
...
+
+            for (ItemStack stack : stacks) {
+                if (!stack.isEmpty()) {
+                    counts.merge(stack.getItem(), stack.getCount(), Integer::sum);
+                }
+            }
+
+            Set<Entry> entries = counts.entrySet().stream()
+                    .map(entry -> new Entry(entry.getKey(), entry.getValue()))
+                    .sorted(Comparator.comparing(entry -> BuiltInRegistries.ITEM.getKey(entry.item()).toString()))
```
Roelymole: I don't think I quite registered this line to sort the entries before, so either it should be removed or we could change back to using a list.

I'm not sure if there is much difference between the two approaches, but my guess would be that sorting is slightly more expensive.

src/main/java/dev/galacticraft/mod/content/item/CannedFoodItem.java
```diff
@@ -272,167 +276,153 @@ public static boolean isCannedFoodItem(ItemStack stack) {
...
-            int blue = color & 0xFF;
+        ItemStack itemStackToAdd = stack.copyWithCount(countToAdd);
+        int existingIndex = -1;
 
-            sumRed += red;
-            sumGreen += green;
-            sumBlue += blue;
-            count++;
+        for (int i = 0; i < cannedFoodItems.size(); i++) {
+            if (ItemStack.isSameItemSameComponents(cannedFoodItems.get(i).copyWithCount(countToAdd), itemStackToAdd)) {
```
Roelymole: `ItemStack.isSameItemSameComponents` doesn't take into account the number of items, so you might want to use `ItemStack.matches` instead if that is what you are after. If not, then the `copyWithCount` seems a bit redundant.

Roelymole (commented):
I have done a fair bit of testing on a dedicated server and haven't run into any major problems.

When you have time it would be good if you could see if you are happy with the changes that I made.

src/main/java/dev/galacticraft/mod/network/c2s/CapeSelectionPayload.java
```diff
@@ -57,17 +61,48 @@ public record CapeSelectionPayload(CapeMode mode, String gcCapeId) implements C2
...
 
-        if (!ServerCapeManager.validateSelection(player, m, id)) {
-            m = CapeMode.VANILLA;
-            id = null;
+        CapeMode mode = this.mode == null ? CapeMode.VANILLA : this.mode;
+        String gcCapeId = this.gcCapeId;
+
+        String playerName = player.getGameProfile().getName();
+        String uuid = player.getUUID().toString();
+        CapeRole role = CapesLoader.roleFor(uuid);
```
Roelymole: These values don't seem to be needed.

maxryan008:
I've been offline for quite a while but I finally have a bit of spare time again so I will try to get around to this sometime this week

src/main/java/dev/galacticraft/mod/content/item/CannedFoodItem.java
maxryan008: Yes copyWithCount is the redundant part I will change it to just copy()

src/main/java/dev/galacticraft/mod/client/util/CannedFoodColorCache.java
maxryan008: done

src/main/java/dev/galacticraft/mod/network/c2s/CapeSelectionPayload.java
maxryan008: Removed

Roelymole (approved):
I don't think there are any blocking issues anymore.

One thing that I was reminded of is the `FabricItemMixin` is client-side only, which might be worth rethinking. It doesn't seem to be breaking anything at the moment, but could be a potential pitfall.

Am I correct in saying that this was so that AppleSkin would display the correct previews?

I haven't been able to think of why we wouldn't want it on the server-side, other than to prevent it from being persisted, so is that main reason?

maxryan008:
> I don't think there are any blocking issues anymore.
> 
> One thing that I was reminded of is the `FabricItemMixin` is client-side only, which might be worth rethinking. It doesn't seem to be breaking anything at the moment, but could be a potential pitfall.
> 
> Am I correct in saying that this was so that AppleSkin would display the correct previews?
> 
> I haven't been able to think of why we wouldn't want it on the server-side, other than to prevent it from being persisted, so is that main reason?

Its been a while since I made this but from memory there is a few reasons its client side. Firstly I believe apple skin is client side and doesn't like server interaction much, but the second main reason is that the way the canned food determines how much food is to be consumed is based on the players current hunger values and I was passing in the player via Minecraft.getInstance().player which is a client side only reference. I presume its possible to make it work on the server instead but it seems like a lot of overhead when apple skin already handles essentially all of its logic client side.

src/main/java/dev/galacticraft/mod/content/item/CannedFoodItem.java
```diff
@@ -95,111 +91,102 @@ public class CannedFoodItem extends Item implements FabricItemStack {
...
+                    foodItem.finishUsingItem(foodItemCopy, level, player);
                 }
+
                 toConsume -= itemCount;
-                if (toConsume == 0) break;
+                if (toConsume == 0) {
+                    break;
+                }
             }
-            super.finishUsingItem(itemStack, level, livingEntity);
```
Roelymole, line 169: I could be wrong, but I don't think we actually need the `FabricItemMixin` after removing this line (since it was behaving differently on the client and the server).

AppleSkin seems to behave correctly without it.

src/main/java/dev/galacticraft/mod/misc/cape/ServerCapeManager.java
```diff
@@ -42,21 +41,45 @@ public Assignment(CapeMode mode, String gcCapeId) {
...
 
-    public static Assignment get(String uuid) { return ACTIVE.get(uuid.toLowerCase(Locale.ROOT)); }
+    public static Assignment get(String uuid) {
+        if (uuid == null) return null;
+        return ACTIVE.get(uuid.toLowerCase(Locale.ROOT));
+    }
 
     public static void set(ServerPlayer player, CapeMode mode, String gcCapeId) {
         String id = player.getUUID().toString().toLowerCase(Locale.ROOT);
```
Roelymole: I have just come across the `getStringUUID()` method in the Entity class, which returns a precomputed value and is slightly more concise too!

There is at least one other instance of `player.getUUID().toString()` in this file, so it would be worth searching for any other places this has been used.

build.gradle.kts
```diff
@@ -161,7 +161,8 @@ loom {
 
 repositories {
     mavenLocal()
-    maven("https://repo.terradevelopment.net/repository/maven-releases/") {
+    maven("http://5.161.199.182/repository/maven-releases/") {
+        isAllowInsecureProtocol = true
```
Roelymole, line 165: The Pull Request Check workflow still seems to be searching for the terradevelopment Maven, so we might need to make this change on the main branch before that check has any hope of passing.

Roelymole: Actually, it appears that this change needs to be made in the `settings.gradle.kts` file as well.

PR #579, opened 2026-06-08
Feature/subtitles

TakeABao23:
I added subtitles for all the sounds currently in GC5.


src/main/java/dev/galacticraft/mod/data/GCTranslationProvider.java
```diff
@@ -911,7 +911,24 @@ protected void generateChatTranslations() {
         this.add(Chat.CANNOT_FEED_IN_NO_ATMOSPHERE, "You can't feed a pet that is holding its breath.");
         this.add(Chat.CANNOT_FEED_WITH_MASK, "You can't feed that to your pet while it is wearing a mask.");
 
+        this.add(Subtitles.PLAYER_UNLOCKCHEST, "Chest unlocks");
+        this.add(Subtitles.PLAYER_PARACHUTE, "Parachute opens");
+        this.add(Subtitles.PLAYER_OPENAIRLOCK, "Airlock opens");
+        this.add(Subtitles.PLAYER_CLOSEAIRLOCK, "Airlock closes");
+        this.add(Subtitles.ENTITY_SLIME_DEATH, "Slimeling dies");
+        this.add(Subtitles.ENTITY_OOH, "ooh");
```
Roelymole: The capitalization of this subtitle is inconsistent.

src/main/java/dev/galacticraft/mod/data/GCTranslationProvider.java
```diff
@@ -911,7 +911,24 @@ protected void generateChatTranslations() {
...
+        this.add(Subtitles.ENTITY_OUCH, "Ouch");
+        this.add(Subtitles.ENTITY_BOSSLAUGH, "Evolved Skeleton Boss laughs");
+        this.add(Subtitles.ENTITY_BOSSDEATH, "Evolved Skeleton Boss dies");
+        this.add(Subtitles.ENTITY_ASTROMINER, "Astrominer flies");
         this.add(Subtitles.THROW_METEOR_CHUNK, "Meteor Chunk flies");
+        this.add(Subtitles.AMBIENCE_SINGLEDRIP, "Liquid drips");
+        this.add(Subtitles.AMBIENCE_SCARYSCAPE, "Eerie Noise");
```
Roelymole: Noise should probably not be capitalized here.

src/main/java/dev/galacticraft/mod/data/GCTranslationProvider.java
Roelymole, line 1: Since there are quite a few subtitles translations now, it might be better to put them in a separate helper method `generateSubtitleTranslations` or `generateSubtitlesTranslations`.

src/main/java/dev/galacticraft/mod/util/Translations.java
```diff
@@ -128,7 +128,24 @@ interface Chat {
...
+        String ENTITY_BOSSDEATH = "subtitles.galacticraft.entity.bossdeath";
+        String ENTITY_ASTROMINER = "subtitles.galacticraft.entity.astrominer";
         String THROW_METEOR_CHUNK = "subtitles.galacticraft.entity.throwable_meteor_chunk.throw";
+        String AMBIENCE_SINGLEDRIP = "subtitles.galacticraft.ambience.singledrip";
+        String AMBIENCE_SCARYSCAPE = "subtitles.galacticraft.ambience.scaryscape";
+        String SHUTTLE_SHUTTLE = "subtitles.galacticraft.shuttle.shuttle";
+        String MACHINE_CIRCUITSCRITCH = "subtitles.galacticraft.machine.circuitscritch";
+        String MACHINE_MACHINEBUZZ = "subtitles.galacticraft.machine.machinebuzz";
+        String MACHINE_MACHINEWHIR = "subtitles.galacticraft.machine.machinewhir";
+        String MACHINE_OXYGENFAN = "subtitles.galacticraft.machine.oxygenfan";
```
Roelymole: I am not really a fan of the naming conventions for the sounds, so now is the best time to get the translation keys right before people start translating the subtitles on Crowdin.

I think we should separate the words with underscores, e.g. `unlock_chest`, `open_airlock`, `boss_laugh` etc.

TakeABao23: Should we do that with `"subtitles.galacticraft.player.unlock_chest"` or just `PLAYER_UNLOCK_CHEST`?

Also, I'm thinking to rename `THROW_METEOR_CHUNK` to either `ENTITY_THROW_METEOR_CHUNK` or `PLAYER_THROW_METEOR_CHUNK` just for consistency's sake. What do you think would be best?

src/main/java/dev/galacticraft/mod/data/GCTranslationProvider.java
```diff
@@ -33,6 +34,32 @@
...
+import dev.galacticraft.mod.util.Translations.Misc;
+import dev.galacticraft.mod.util.Translations.RecipeCategory;
+import dev.galacticraft.mod.util.Translations.RegistryDebug;
+import dev.galacticraft.mod.util.Translations.SetOxygen;
+import dev.galacticraft.mod.util.Translations.SolarPanel;
+import dev.galacticraft.mod.util.Translations.SpaceRace;
+import dev.galacticraft.mod.util.Translations.Subtitles;
+import dev.galacticraft.mod.util.Translations.Tooltip;
+import dev.galacticraft.mod.util.Translations.Ui;
+import dev.galacticraft.mod.util.Translations.Waila;
```
Roelymole: There is already a static import of all these.

src/main/java/dev/galacticraft/mod/data/GCTranslationProvider.java
```diff
@@ -24,6 +24,7 @@
 
 import dev.galacticraft.api.data.TranslationProvider;
 import dev.galacticraft.api.rocket.part.RocketPart;
+import dev.galacticraft.mod.Constant.Cape;
```
Roelymole: Similarly, there is already a static import of this.

src/main/java/dev/galacticraft/mod/util/Translations.java
Roelymole: The main thing is to get the strings themselves correct (e.g. `"subtitles.galacticraft.player.unlock_chest"`) as changing these later would have more of an impact, whereas the variable names don't matter so much.

If anything I would question the need for the `PLAYER_`, `ENTITY_`, `SHUTTLE_` and `MACHINE_` prefixes, because they seem quite redundant.

src/main/java/dev/galacticraft/mod/data/GCTranslationProvider.java
TakeABao23: grrr, my IDE keeps automatically adding them for some reason

Roelymole: You could maybe make it a non-static wildcard import if it is still causing problems.

```java
import dev.galacticraft.mod.util.Translations.*;
```

src/main/resources/assets/galacticraft/sounds.json
```diff
@@ -63,182 +63,199 @@
...
+        "name": "galacticraft:machine/oxygen/fan",
         "stream": false
       }
     ]
   },
   "music_disc.legacy.mars": {
     "sounds": [
       {
-        "name": "galacticraft:music/legacy/mars",
+        "name": "galacticraft:music/records/legacy/mars",
```
Roelymole: If we are following the precedence of vanilla, this would not be nested inside the `music` folder, i.e. just `"galacticraft:records/legacy/mars"`

src/main/java/dev/galacticraft/mod/content/GCSounds.java
```diff
@@ -30,59 +30,61 @@
...
-    public static final Holder.Reference<SoundEvent> MUSIC_MOON = register("music.moon");
+    // Ambient
+    public static final SoundEvent SCARYSCAPE = SoundEvent.createVariableRangeEvent(Constant.id("ambient.scaryscape"));
+    public static final SoundEvent SINGLE_DRIP = SoundEvent.createVariableRangeEvent(Constant.id("ambient.single_drip"));
+    // Entity
+    public static final SoundEvent ASTROMINER = SoundEvent.createVariableRangeEvent(Constant.id("entity.astrominer"));
+    public static final SoundEvent BOSS_DEATH = SoundEvent.createVariableRangeEvent(Constant.id("entity.boss.death"));
+    public static final SoundEvent BOSS_LAUGH = SoundEvent.createVariableRangeEvent(Constant.id("entity.boss.laugh"));
+    public static final SoundEvent BOSS_OOH = SoundEvent.createVariableRangeEvent(Constant.id("entity.boss.ooh"));
+    public static final SoundEvent BOSS_OUCH = SoundEvent.createVariableRangeEvent(Constant.id("entity.boss.ouch"));
```
Roelymole: It might be good to go with the prefix `SKELETON_BOSS_` ready for when other bosses are added.

May you also make similar changes in all the other places (you know the drill).

src/main/java/dev/galacticraft/mod/content/GCSounds.java
```diff
@@ -30,59 +30,61 @@
...
+    public static final SoundEvent METEOR_THROW = SoundEvent.createVariableRangeEvent(Constant.id("entity.meteor_chunk.throw"));
+    public static final SoundEvent ROCKET_IGNITE = SoundEvent.createVariableRangeEvent(Constant.id("entity.rocket.ignite"));
+    public static final SoundEvent SLIME_DEATH = SoundEvent.createVariableRangeEvent(Constant.id("entity.slime.death"));
```
Roelymole: Please could you change this from `SLIME_DEATH` to `SLIMELING_DEATH`, because the former is more ambiguous.

Again, `slime` should be replaced with `slimeling` in all the other places too.

src/main/java/dev/galacticraft/mod/content/GCSounds.java
```diff
@@ -30,59 +30,61 @@
 
 public class GCSounds {
     private static final GCRegistry<SoundEvent> SOUNDS = new GCRegistry<>(BuiltInRegistries.SOUND_EVENT);
-    public static final Holder.Reference<SoundEvent> MUSIC_MOON = register("music.moon");
+    // Ambient
+    public static final SoundEvent SCARYSCAPE = SoundEvent.createVariableRangeEvent(Constant.id("ambient.scaryscape"));
+    public static final SoundEvent SINGLE_DRIP = SoundEvent.createVariableRangeEvent(Constant.id("ambient.single_drip"));
```
Roelymole: It would be more consistent with vanilla to use the `AMBIENT_` prefix for these.

src/main/java/dev/galacticraft/mod/content/GCSounds.java
```diff
@@ -30,59 +30,61 @@
...
     public static final Holder.Reference<SoundEvent> MUSIC_LEGACY_MIMAS = register("music_disc.legacy.mimas");
     public static final Holder.Reference<SoundEvent> MUSIC_LEGACY_ORBIT = register("music_disc.legacy.orbit");
     public static final Holder.Reference<SoundEvent> MUSIC_LEGACY_SPACERACE = register("music_disc.legacy.spacerace");
+    // Player
+    public static final SoundEvent CHEST_UNLOCK = SoundEvent.createVariableRangeEvent(Constant.id("player.chest.unlock"));
+    public static final SoundEvent PARACHUTE = SoundEvent.createVariableRangeEvent(Constant.id("player.parachute"));
 
     public static Holder.Reference<SoundEvent> register(String id) {
         return SOUNDS.registerForHolder(id, SoundEvent.createVariableRangeEvent(Constant.id(id)));
     }
```
Roelymole: I think this should be made private and renamed to `registerForHolder`.

A couple more helper method should be added for registering a `SoundEvent` without a holder, because there is a lot of repeated code.

```java
private static SoundEvent register(String id) {
    return register(Constant.id(id));
}

private static SoundEvent register(ResourceLocation id) {
    return Registry.register(BuiltInRegistries.SOUND_EVENT, id, SoundEvent.createVariableRangeEvent(id));
}
```

I think we will need to still call `GCSounds.register();` in the mod initializer (even if it is empty) to avoid intermittent errors about unregistered holders.

src/main/java/dev/galacticraft/mod/content/GCSounds.java
```diff
@@ -26,63 +26,55 @@
...
+    public static final SoundEvent ROCKET_IGNITE = register("entity.rocket.ignite");
+    public static final SoundEvent SLIMELING_DEATH = register("entity.slimeling.death");
     // Machine Sounds
-    public static final SoundEvent CIRCUIT_SCRITCH = SoundEvent.createVariableRangeEvent(Constant.id("machine.circuitscritch"));
-    public static final SoundEvent MACHINE_BUZZ = SoundEvent.createVariableRangeEvent(Constant.id("machine.machinebuzz"));
-    public static final SoundEvent MACHINE_HUM = SoundEvent.createVariableRangeEvent(Constant.id("machine.machinehum"));
-    public static final SoundEvent OXYGEN_FAN = SoundEvent.createVariableRangeEvent(Constant.id("machine.oxygenfan"));
-
+    public static final SoundEvent AIRLOCK_OPEN = register("machine.airlock.open");
+    public static final SoundEvent AIRLOCK_CLOSE = register("machine.airlock.close");
```
Roelymole: These two are the wrong way round if you were ordering them alphabetically within each group.

Roelymole (approved):
Thanks for the good work!

PR #582, opened 2026-07-13
Fix: make rocket ignition sound play immediately

TakeABao23:


Galacticraft/src/main/resources/assets/galacticraft/sounds/entity/rocket/ignite.ogg
Roelymole, line 1: The size of this sound file seems to have more than doubled. It might be worth using the same level of compression as long as that doesn't seem to be part of the issue before.

Roelymole (approved):
Thanks!

PR #583, opened 2026-07-13
Refactor/entity folders

TakeABao23:
Reorganized entity folders and fixed an evolved creeper sound bug


TakeABao23:
down daemon = failed checks..?

Roelymole (approved):
None of these changes are controversial at all.

The only thing to really consider is whether any of the entities being moved to the `moon` subfolder might be found in other dimensions, but that is a bridge we can cross when we come to it.

PR #588, opened 2026-09-07
Fix glitching items when dimensions change

### LINKED ISSUE #580: Galacticraft Inv Clearing When Using Rocket
### Version Information

40f450b

### Log or Crash Report

_No response_

### Reproduction steps

When you travel to another dimension via a rocket, the custom inventory(the menu with oxygen tanks, thermal gear, etc.) is cleared, but when you click on a slot, it returns to its previous state

emsar69:
Title.

After player changes dimensions using rocket the custom inventory being reinitialized with no items. Sending proper [GearInvPayload.java](https://github.com/TeamGalacticraft/Galacticraft/blob/main/src/main/java/dev/galacticraft/impl/network/s2c/GearInvPayload.java) packet to player solves the bug.

Closes #580

### DIFF
--- a/src/main/java/dev/galacticraft/impl/internal/fabric/GalacticraftAPI.java
+++ b/src/main/java/dev/galacticraft/impl/internal/fabric/GalacticraftAPI.java
@@ -36,18 +36,23 @@ import dev.galacticraft.dynamicdimensions.api.event.DynamicDimensionLoadCallback
 import dev.galacticraft.impl.internal.command.GCApiCommands;
 import dev.galacticraft.impl.network.GCApiPackets;
 import dev.galacticraft.impl.network.GCApiServerPacketReceivers;
+import dev.galacticraft.impl.network.s2c.GearInvPayload;
 import dev.galacticraft.impl.universe.BuiltinObjects;
 import dev.galacticraft.mod.Constant;
 import dev.galacticraft.mod.data.gen.SatelliteChunkGenerator;
 import net.fabricmc.api.ModInitializer;
+import net.fabricmc.fabric.api.entity.event.v1.ServerEntityWorldChangeEvents;
 import net.fabricmc.fabric.api.event.lifecycle.v1.ServerLifecycleEvents;
 import net.fabricmc.fabric.api.event.registry.DynamicRegistries;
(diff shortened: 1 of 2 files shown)


src/main/java/dev/galacticraft/impl/internal/fabric/GalacticraftAPI.java
Roelymole, line 1: It would make more sense for this code to be in [GCEventHandlers.java](https://github.com/TeamGalacticraft/Galacticraft/blob/main/src/main/java/dev/galacticraft/mod/events/GCEventHandlers.java)

src/main/java/dev/galacticraft/impl/internal/fabric/GalacticraftAPI.java
```diff
@@ -86,6 +91,16 @@ public void onInitialize() {
             ((SatelliteAccessor) minecraftServer).galacticraft$loadSatellites(dynamicDimensionLoader);
         });
 
+        ServerEntityWorldChangeEvents.AFTER_PLAYER_CHANGE_WORLD.register(((player, origin, destination) -> {
+            Container inv = player.galacticraft$getGearInv();
+            ItemStack[] stacks = new ItemStack[inv.getContainerSize()];
+            for(int i = 0; i < inv.getContainerSize(); i++) {
```
Roelymole: Please could you put a space between the `for` and the open bracket.

Roelymole (commented):
Thanks for fixing this issue, I have a couple of comments, but they should be fairly quick to address.

Please could you also merge the latest changes from the main branch so that I can run the workflow.

Roelymole (commented):
I think you might still need to use `ServerEntityWorldChangeEvents.AFTER_PLAYER_CHANGE_WORLD` otherwise this issue will occur when going to the nether, for example.

emsar69:
> I think you might still need to use `ServerEntityWorldChangeEvents.AFTER_PLAYER_CHANGE_WORLD` otherwise this issue will occur when going to the nether, for example.

Yeah I've missed that thanks. I tested it in both dimensions and planets. It's working as intended.

Roelymole (approved):

PR #589, opened 2026-09-08
Updated Fallen Meteor Block

MaverXka:
New knockback behavior.
Cooler faster when in water.
Hot meteor not affecting player in creative and entity with fire immune or fire resistant

### DIFF
--- a/src/main/java/dev/galacticraft/mod/content/block/environment/FallenMeteorBlock.java
+++ b/src/main/java/dev/galacticraft/mod/content/block/environment/FallenMeteorBlock.java
@@ -23,7 +23,6 @@
 package dev.galacticraft.mod.content.block.environment;
 
 import com.mojang.serialization.MapCodec;
-import dev.galacticraft.mod.Constant;
 import net.minecraft.core.BlockPos;
 import net.minecraft.core.Direction;
 import net.minecraft.core.particles.ParticleTypes;
--- a/src/main/java/dev/galacticraft/mod/content/block/environment/FallenMeteorBlock.java
+++ b/src/main/java/dev/galacticraft/mod/content/block/environment/FallenMeteorBlock.java
@@ -246,7 +246,7 @@ public class FallenMeteorBlock extends FallingBlock implements SimpleWaterlogged
     public void onLand(Level level, BlockPos pos, BlockState fallingBlockState, BlockState currentStateInPos, FallingBlockEntity blockEntity) {
         super.onLand(level, pos, fallingBlockState, currentStateInPos, blockEntity);
 
-        for (var livingEntity : level.getEntitiesOfClass(LivingEntity.class, blockEntity.getBoundingBox().inflate(1.0D)))
+        for (var livingEntity : level.getEntitiesOfClass(LivingEntity.class, blockEntity.getBoundingBox()))
         {
             var damage = Math.max(2.0F, Math.min(blockEntity.fallDistance * 2.0F, 40.0F));
             livingEntity.hurt(level.damageSources().fallingBlock(blockEntity), damage);
--- a/src/main/java/dev/galacticraft/mod/content/block/environment/FallenMeteorBlock.java
+++ b/src/main/java/dev/galacticraft/mod/content/block/environment/FallenMeteorBlock.java
@@ -34,8 +34,12 @@ import net.minecraft.tags.FluidTags;
 import net.minecraft.util.FastColor;
 import net.minecraft.util.Mth;
 import net.minecraft.util.RandomSource;
+import net.minecraft.world.effect.MobEffects;
 import net.minecraft.world.entity.Entity;
 import net.minecraft.world.entity.LivingEntity;
+import net.minecraft.world.entity.ai.attributes.Attributes;
+import net.minecraft.world.entity.item.FallingBlockEntity;
+import net.minecraft.world.entity.player.Player;
 import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.item.context.BlockPlaceContext;
 import net.minecraft.world.item.enchantment.EnchantmentHelper;
@@ -55,6 +59,7 @@ import net.minecraft.world.level.block.state.properties.BooleanProperty;
 import net.minecraft.world.level.block.state.properties.IntegerProperty;
 import net.minecraft.world.level.material.FluidState;
 import net.minecraft.world.level.material.Fluids;
+import net.minecraft.world.phys.Vec3;
 import net.minecraft.world.phys.shapes.CollisionContext;
 import net.minecraft.world.phys.shapes.VoxelShape;
 import org.jetbrains.annotations.Nullable;
@@ -64,6 +69,7 @@ public class FallenMeteorBlock extends FallingBlock implements SimpleWaterlogged
     private static final VoxelShape SHAPE = box(3, 0, 3, 13, 10, 13);
     public static final BooleanProperty WATERLOGGED = BlockStateProperties.WATERLOGGED;
     public static final IntegerProperty HEAT = IntegerProperty.create("heat", 0, 5);
+    private final int[] coolingRates = { 1, 500, 460, 370, 240, 200 };
 
     public FallenMeteorBlock(BlockBehaviour.Properties settings) {
         super(settings);
@@ -83,14 +89,58 @@ public class FallenMeteorBlock extends FallingBlock implements SimpleWaterlogged
     @Override
     public void tick(BlockState state, ServerLevel world, BlockPos pos, RandomSource random) {
         int i = state.getValue(HEAT);
+        boolean waterlogged = state.getValue(WATERLOGGED);
 
         if (i > 0) {
-            if (random.nextInt(500) == 0) {
+            int coolingRate = coolingRates[i];
+
+            if (waterlogged)
+            {
+                coolingRate /= 2;
+            }
+
+            if (random.nextInt(coolingRate) == 0) {
                 world.setBlock(pos, state.setValue(HEAT, i - 1), Block.UPDATE_CLIENTS);
+                world.playSound(null,
+                        pos,
+                        SoundEvents.FIRE_EXTINGUISH,
+                        SoundSource.BLOCKS,
+                        0.7F,
+                        0.8F);
             } else {
                 super.tick(state, world, pos, random);
             }
         }
+    }
+
+    @Override
+    public void animateTick(BlockState state, Level level, BlockPos pos, RandomSource random) {
+
+        int heat = state.getValue(HEAT);
+        if ( !(state.getValue(WATERLOGGED) && heat > 0) ) {
+            return;
+        }
+
+        double hspread = 0;
+        if (heat > 3) {
+            double spread = (heat - 3) * 0.03D;
+
+            hspread = (random.nextDouble() - 0.5D) * spread;
+        }
+
+        for (var i = 0; i < heat; ++i) {
+            level.addParticle(
+                    ParticleTypes.CLOUD,
+                    pos.getX() + random.nextDouble(),
+                    pos.getY() + 0.2D + random.nextDouble(),
+                    pos.getZ() + random.nextDouble(),
+                    hspread,
+                    heat * 0.01D,
+                    hspread
+            );
+        }
+
+
     }
 
     @Override
@@ -129,7 +179,17 @@ public class FallenMeteorBlock extends FallingBlock implements SimpleWaterlogged
             return;
         }
 
+        if (entity instanceof Player player)
+        {
+            if (player.isCreative()) return;
+        }
+
         if (entity instanceof LivingEntity livingEntity) {
+
+            if (entity.fireImmune() || livingEntity.hasEffect(MobEffects.FIRE_RESISTANCE)) {
+                return;
+            }
+
             world.playSound(null, pos.getX() + 0.5F, pos.getY() + 0.5F, pos.getZ() + 0.5F, SoundEvents.GENERIC_EXTINGUISH_FIRE, SoundSource.NEUTRAL, 0.5F, 2.6F + (world.random.nextFloat() - world.random.nextFloat()) * 0.8F);
 
             for (var i = 0; i < 8; ++i) {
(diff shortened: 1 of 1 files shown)


src/main/java/dev/galacticraft/mod/content/block/environment/FallenMeteorBlock.java
```diff
@@ -129,25 +179,77 @@ public void entityInside(BlockState state, Level world, BlockPos pos, Entity ent
             return;
         }
 
+        if (entity instanceof Player player)
+        {
```
Roelymole: Please could you follow the style of putting the opening curly brackets on the same line.

```suggestion
        if (entity instanceof Player player) {
```

src/main/java/dev/galacticraft/mod/content/block/environment/FallenMeteorBlock.java
```diff
@@ -83,16 +89,60 @@ public void randomTick(BlockState state, ServerLevel world, BlockPos pos, Random
...
     public void tick(BlockState state, ServerLevel world, BlockPos pos, RandomSource random) {
         int i = state.getValue(HEAT);
+        boolean waterlogged = state.getValue(WATERLOGGED);
 
         if (i > 0) {
-            if (random.nextInt(500) == 0) {
+            int coolingRate = coolingRates[i];
+
+            if (waterlogged)
+            {
```
Roelymole: Please could you follow the style of putting the opening curly brackets on the same line.

```suggestion
            if (waterlogged) {
```

src/main/java/dev/galacticraft/mod/content/block/environment/FallenMeteorBlock.java
```diff
@@ -129,25 +179,77 @@ public void entityInside(BlockState state, Level world, BlockPos pos, Entity ent
...
+                return;
+            }
+
             world.playSound(null, pos.getX() + 0.5F, pos.getY() + 0.5F, pos.getZ() + 0.5F, SoundEvents.GENERIC_EXTINGUISH_FIRE, SoundSource.NEUTRAL, 0.5F, 2.6F + (world.random.nextFloat() - world.random.nextFloat()) * 0.8F);
 
             for (var i = 0; i < 8; ++i) {
                 world.addParticle(ParticleTypes.LARGE_SMOKE, pos.getX() + Math.random(), pos.getY() + 0.2D + Math.random(), pos.getZ() + Math.random(), 0.0D, 0.0D, 0.0D);
             }
 
             if (!livingEntity.isOnFire()) {
```
Roelymole: I don't think we need to check if the entity is already on fire, since `igniteForSeconds` does not decrease the time that the entity is on fire for, unlike `setRemainingFireTicks`.

src/main/java/dev/galacticraft/mod/content/block/environment/FallenMeteorBlock.java
```diff
@@ -129,25 +179,77 @@ public void entityInside(BlockState state, Level world, BlockPos pos, Entity ent
...
+        );
+    }
+
+    @Override
+    public void onLand(Level level, BlockPos pos, BlockState fallingBlockState, BlockState currentStateInPos, FallingBlockEntity blockEntity) {
+        super.onLand(level, pos, fallingBlockState, currentStateInPos, blockEntity);
+
+        for (var livingEntity : level.getEntitiesOfClass(LivingEntity.class, blockEntity.getBoundingBox().inflate(1.0D)))
+        {
+            var damage = Math.max(2.0F, Math.min(blockEntity.fallDistance * 2.0F, 40.0F));
```
Roelymole: Using `Mth.clamp(value, min, max)` is more concise.

```suggestion
            float damage = Mth.clamp(blockEntity.fallDistance * 2.0F, 2.0F, 40.0F);
```

src/main/java/dev/galacticraft/mod/content/block/environment/FallenMeteorBlock.java
```diff
@@ -83,16 +89,60 @@ public void randomTick(BlockState state, ServerLevel world, BlockPos pos, Random
...
+            if (random.nextInt(coolingRate) == 0) {
                 world.setBlock(pos, state.setValue(HEAT, i - 1), Block.UPDATE_CLIENTS);
+                world.playSound(null,
+                        pos,
+                        SoundEvents.FIRE_EXTINGUISH,
+                        SoundSource.BLOCKS,
+                        0.7F,
+                        0.8F);
             } else {
                 super.tick(state, world, pos, random);
```
Roelymole: The super method needs to be called for fallen meteors with heat level 0 so that they will also fall if there is no block beneath them.

src/main/java/dev/galacticraft/mod/content/block/environment/FallenMeteorBlock.java
```diff
@@ -83,16 +89,60 @@ public void randomTick(BlockState state, ServerLevel world, BlockPos pos, Random
...
             }
         }
     }
 
+    @Override
+    public void animateTick(BlockState state, Level level, BlockPos pos, RandomSource random) {
```
Roelymole, line 112: I think randomly playing some kind of sound while the fallen meteor is cooling down in water would be good, rather than just when the heat level changes.

src/main/java/dev/galacticraft/mod/content/block/environment/FallenMeteorBlock.java
```diff
@@ -129,25 +179,77 @@ public void entityInside(BlockState state, Level world, BlockPos pos, Entity ent
...
+                velocity.y / 2.0D + direction.y,
+                velocity.z / 2.0D + direction.z
+        );
+    }
+
+    @Override
+    public void onLand(Level level, BlockPos pos, BlockState fallingBlockState, BlockState currentStateInPos, FallingBlockEntity blockEntity) {
+        super.onLand(level, pos, fallingBlockState, currentStateInPos, blockEntity);
+
+        for (var livingEntity : level.getEntitiesOfClass(LivingEntity.class, blockEntity.getBoundingBox().inflate(1.0D)))
```
Roelymole: This seems like a lot to inflate the bounding box by.

Also, please could you put the curly bracket on this line, rather than the next line.

src/main/java/dev/galacticraft/mod/content/block/environment/FallenMeteorBlock.java
```diff
@@ -129,25 +179,77 @@ public void entityInside(BlockState state, Level world, BlockPos pos, Entity ent
...
+
+        Vec3 velocity = entity.getDeltaMovement();
+
+        entity.hasImpulse = true;
+
+        entity.setDeltaMovement(
+                velocity.x / 2.0D + direction.x,
+                velocity.y / 2.0D + direction.y,
+                velocity.z / 2.0D + direction.z
+        );
```
Roelymole, line 237: This can be made more concise by doing something like the following:

```suggestion
        entity.setDeltaMovement(
                velocity.scale(0.5D).add(direction)
        );
```

src/main/java/dev/galacticraft/mod/content/block/environment/FallenMeteorBlock.java
```diff
@@ -129,25 +179,77 @@ public void entityInside(BlockState state, Level world, BlockPos pos, Entity ent
...
+                livingEntity.igniteForSeconds(1);
             }
 
-            var knockX = pos.getX() + 0.5F - livingEntity.getX();
+            var knockX = livingEntity.getX() - (pos.getX());
+            var knockY = livingEntity.getY() + livingEntity.getBbHeight() / 2.0D - (pos.getY());
             double knockZ;
 
-            for (knockZ = livingEntity.getZ() - pos.getZ(); knockX * knockX + knockZ * knockZ < 1.0E-4D; knockZ = (Math.random() - Math.random()) * 0.01D) {
+            for (knockZ = livingEntity.getZ() - (pos.getZ()); knockX * knockX + knockZ * knockZ < 1.0E-4D; knockZ = (Math.random() - Math.random()) * 0.01D) {
```
Roelymole, line 204: You seem to have added in some unnecessary brackets around `pos.getX()`,  `pos.getY()` and  `pos.getZ()`.

src/main/java/dev/galacticraft/mod/content/block/environment/FallenMeteorBlock.java
```diff
@@ -129,25 +179,77 @@ public void entityInside(BlockState state, Level world, BlockPos pos, Entity ent
...
                 knockX = (Math.random() - Math.random()) * 0.01D;
             }
 
-            livingEntity.knockback(1, knockX, knockZ);
+            meteorKnockback(livingEntity,0.3F, knockX, knockY, knockZ);
```
Roelymole: Please could you add a space after the first comma.

```suggestion
            meteorKnockback(livingEntity, 0.3F, knockX, knockY, knockZ);
```

src/main/java/dev/galacticraft/mod/content/block/environment/FallenMeteorBlock.java
```diff
@@ -129,25 +179,77 @@ public void entityInside(BlockState state, Level world, BlockPos pos, Entity ent
...
+                velocity.z / 2.0D + direction.z
+        );
+    }
+
+    @Override
+    public void onLand(Level level, BlockPos pos, BlockState fallingBlockState, BlockState currentStateInPos, FallingBlockEntity blockEntity) {
+        super.onLand(level, pos, fallingBlockState, currentStateInPos, blockEntity);
+
+        for (var livingEntity : level.getEntitiesOfClass(LivingEntity.class, blockEntity.getBoundingBox()))
+        {
```
Roelymole: Please could you follow the style of putting opening curly brackets on the same line.

src/main/java/dev/galacticraft/mod/content/block/environment/FallenMeteorBlock.java
```diff
@@ -83,16 +89,60 @@ public void randomTick(BlockState state, ServerLevel world, BlockPos pos, Random
...
+                world.playSound(null,
+                        pos,
+                        SoundEvents.FIRE_EXTINGUISH,
+                        SoundSource.BLOCKS,
+                        0.7F,
+                        0.8F);
             } else {
                 super.tick(state, world, pos, random);
             }
         }
```
Roelymole: The super method needs to be called for fallen meteors with heat level 0 so that they will also fall if there is no block beneath them.

I would suggest adding a return statement after playing the sound and then move the call to the super method outside of the if/else blocks.

```suggestion
                world.playSound(
                        null,
                        pos,
                        SoundEvents.FIRE_EXTINGUISH,
                        SoundSource.BLOCKS,
                        0.7F,
                        0.8F
                );

                return;
            }
        }

        super.tick(state, world, pos, random);
```

src/main/java/dev/galacticraft/mod/content/block/environment/FallenMeteorBlock.java
```diff
@@ -129,25 +179,77 @@ public void entityInside(BlockState state, Level world, BlockPos pos, Entity ent
...
+    @Override
+    public void onLand(Level level, BlockPos pos, BlockState fallingBlockState, BlockState currentStateInPos, FallingBlockEntity blockEntity) {
+        super.onLand(level, pos, fallingBlockState, currentStateInPos, blockEntity);
+
+        for (var livingEntity : level.getEntitiesOfClass(LivingEntity.class, blockEntity.getBoundingBox()))
+        {
+            var damage = Math.max(2.0F, Math.min(blockEntity.fallDistance * 2.0F, 40.0F));
+            livingEntity.hurt(level.damageSources().fallingBlock(blockEntity), damage);
         }
     }
```
Roelymole: This does nowhere near as much damage as an anvil, which uses the following approach.

```java
	private static final float FALL_DAMAGE_PER_DISTANCE = 2.0F;
	private static final int FALL_DAMAGE_MAX = 40;

    ...

    @Override
    protected void falling(FallingBlockEntity entity) {
        entity.setHurtsEntities(FALL_DAMAGE_PER_DISTANCE, FALL_DAMAGE_MAX);
    }
```

I think the `onLand` method should be used to play a sound, maybe some kind of heavy thump.

src/main/java/dev/galacticraft/mod/content/block/environment/FallenMeteorBlock.java
```diff
@@ -64,6 +69,7 @@ public class FallenMeteorBlock extends FallingBlock implements SimpleWaterlogged
     private static final VoxelShape SHAPE = box(3, 0, 3, 13, 10, 13);
     public static final BooleanProperty WATERLOGGED = BlockStateProperties.WATERLOGGED;
     public static final IntegerProperty HEAT = IntegerProperty.create("heat", 0, 5);
+    private final int[] coolingRates = { 1, 500, 460, 370, 240, 200 };
```
Roelymole: These values should be much smaller so that it doesn't take absolutely ages for the fallen meteor blocks to cool down!

See the following for the maths: https://minecraft.wiki/w/Random_Tick#Time_estimation

This is the sort of scale I mean:

```suggestion
    private final int[] coolingRates = { 0, 10, 8, 6, 4, 2 };
```

src/main/java/dev/galacticraft/mod/content/block/environment/FallenMeteorBlock.java
```diff
@@ -83,16 +89,60 @@ public void randomTick(BlockState state, ServerLevel world, BlockPos pos, Random
...
+
+        for (var i = 0; i < heat; ++i) {
+            level.addParticle(
+                    ParticleTypes.CLOUD,
+                    pos.getX() + random.nextDouble(),
+                    pos.getY() + 0.2D + random.nextDouble(),
+                    pos.getZ() + random.nextDouble(),
+                    hspread,
+                    heat * 0.01D,
+                    hspread
```
Roelymole: The `velocityX` and `velocityZ` arguments should be independent of each other to avoid the particles all travelling in one of two directions when viewed from above.

src/main/java/dev/galacticraft/mod/content/block/environment/FallenMeteorBlock.java
```diff
@@ -83,16 +89,60 @@ public void randomTick(BlockState state, ServerLevel world, BlockPos pos, Random
     @Override
     public void tick(BlockState state, ServerLevel world, BlockPos pos, RandomSource random) {
```
Roelymole: I am aware that you didn't write this code originally, but I don't think we should be overriding the `tick` method at all.

Instead, this code should be put directly in the `randomTick` method and the `isRandomlyTicking` method should be overridden rather than using the `if (i > 0)` statement to check that the heat level is not zero.

Also, it might be more intuitive if we used `random.nextInt(coolingRate - 1) == 0` rather than `random.nextInt(coolingRate) == 0`, so that the probability that the heat level decreases each time the `randomTick` method is called is `1 / coolingRate` rather than  `1 / (coolingRate + 1)`.

```java
    @Override
    protected boolean isRandomlyTicking(BlockState state) {
        return state.getValue(HEAT) > 0;
    }

    @Override
    public void randomTick(BlockState state, ServerLevel world, BlockPos pos, RandomSource random) {
        int i = state.getValue(HEAT);

        int coolingRate = coolingRates[i];

        if (state.getValue(WATERLOGGED)) {
            coolingRate /= 2;
        }

        if (random.nextInt(coolingRate - 1) == 0) {
            world.setBlock(pos, state.setValue(HEAT, i - 1), Block.UPDATE_CLIENTS);
            world.playSound(
                    null,
                    pos,
                    SoundEvents.FIRE_EXTINGUISH,
                    SoundSource.BLOCKS,
                    0.7F,
                    0.8F
            );
        }
    }
```

Roelymole (changes requested):
The knockback is much more balanced, thanks!

A few more problems have come to my attention from reading through this, the main one being that the fallen meteors are currently taking a ridiculously long time to cool down (plus a misuse of the `tick` method).

src/main/java/dev/galacticraft/mod/content/block/environment/FallenMeteorBlock.java
```diff
@@ -64,6 +69,9 @@ public class FallenMeteorBlock extends FallingBlock implements SimpleWaterlogged
     private static final VoxelShape SHAPE = box(3, 0, 3, 13, 10, 13);
     public static final BooleanProperty WATERLOGGED = BlockStateProperties.WATERLOGGED;
     public static final IntegerProperty HEAT = IntegerProperty.create("heat", 0, 5);
+    private static final float FALL_DAMAGE_PER_DISTANCE = 2.0F;
+    private static final int FALL_DAMAGE_MAX = 40;
+    private final int[] coolingRates = { 0, 10, 8, 6, 4, 2 };
```
Roelymole: This can be made static as well.

It would be good to change the name to `COOLING_RATES` to reflect that these values are constants as well.

src/main/java/dev/galacticraft/mod/content/block/environment/FallenMeteorBlock.java
```diff
@@ -76,21 +84,63 @@ protected MapCodec<FallenMeteorBlock> codec() {
...
+                    pos.getX() + random.nextDouble(),
+                    pos.getY() + 0.2D + random.nextDouble(),
+                    pos.getZ() + random.nextDouble(),
+                    (random.nextDouble() - 0.5D) * spread,
+                    heat * 0.01D,
+                    (random.nextDouble() - 0.5D) * spread
+            );
+        }
+
+        if (level.getGameTime() % 20 == 0) {
```
Roelymole: This does seem somewhat random, but not very natural.

Perhaps using `random.nextInt` here would give a better effect.

src/main/java/dev/galacticraft/mod/content/block/environment/FallenMeteorBlock.java
```diff
@@ -76,21 +84,63 @@ protected MapCodec<FallenMeteorBlock> codec() {
...
+            level.playLocalSound(
+                    pos,
+                    SoundEvents.FIRE_EXTINGUISH,
+                    SoundSource.BLOCKS,
+                    0.7F,
```
Roelymole: Adding some attenuation in volume as the heat level decreases could be a nice feature.

Roelymole (commented):
Just a couple more things.

Also, did you see my comment about playing a heavy thump sound as part of the `onLand` method?

Roelymole (approved):
Thank you for making those changes!

PR #592, opened 2026-09-12
bug fix: coal generator from burning a new fuel when capacitor is full

MaverXka:


src/main/java/dev/galacticraft/mod/content/block/entity/machine/CoalGeneratorBlockEntity.java
```diff
@@ -156,6 +156,11 @@ protected void tickConstant(@NotNull ServerLevel level, @NotNull BlockPos pos, @
         this.energyStorage().insert((long) (Galacticraft.CONFIG.coalGeneratorEnergyProductionRate() * this.heat));
         this.energySource.trySpreadEnergy(level, pos, state);
         profiler.popPush("fuel_reset");
```
Roelymole: This should be split into a `profiler.pop()` here and then `profiler.push("fuel_reset")` after the if statement.

This is to prevent an unbalanced number of profiler pushes/pops due to the return statement.

src/main/java/dev/galacticraft/mod/content/block/entity/machine/CoalGeneratorBlockEntity.java
```diff
@@ -153,25 +153,35 @@ protected void tickConstant(@NotNull ServerLevel level, @NotNull BlockPos pos, @
...
+
         if (this.fuelLength == 0) {
             MachineStatus status = this.consumeFuel(level, pos, state);
             if (status != null) {
+                profiler.pop();
                 return status;
             }
         }
+
         if (++this.fuelTime >= this.fuelLength) {
```
Roelymole: I'm trying to remember what I was thinking when I worked on this before, but can't imagine we want the `consumeFuel` method being called twice in one tick.

It might be impossible for that to happen already, but it would be much more obvious if this was an else if statement.

Roelymole (approved):
Thanks for balancing the profilers.

There is one tiny thing that you have drawn my attention to, but then I'm happy.

Roelymole (approved):
I am pretty convinced now that what it was doing before was only burning the first coal for 319 ticks before starting to burn the second coal on the 320th tick.

Now the first coal burns for the full amount (320 ticks) before the second coal starts burning, which means that the `workTime` in the test should be increased to 321.

PR #593, opened 2026-09-12
feat: add oxygen detector

### LINKED ISSUE #511: Implement the Oxygen Detector
### Version Information

1773de1bb0924feb673fdcfe7b81a8e3e4cf2e54

### Log or Crash Report

_No response_

### Reproduction steps

A good task for someone looking to contribute to the development of Galacticraft 5 would be to implement the Oxygen Detector.

Quite simply, this block should output a redstone signal when there is oxygen present at one of the positions adjacent to the Oxygen Detector and would primarily be used in conjuction with Oxygen Sealers and Oxygen Bubble Distributors.

A slight enhancement on the Oxygen Detector in Galacticraft-Legacy would be to use the side config tab (or similar) to select which directions should be checked for oxygen. This should swap between a texture with a grating to let the oxygen in and a plain face.

Image

emsar69:
Closes #511 

I've implemented oxygen detector. ~~It still needs some small touches, I am creating this PR anyway because It's **mostly** done and for reviews ofc.~~ It's done now.

As mentioned in #511 configuration gui needs machinelib to be updated. I've overrode `cycleFaceConfig` to get around this but It's a temporary solution. See `OxygenDetectorMenu`, also see [MachineLib#27](https://github.com/TeamGalacticraft/MachineLib/pull/27) for related PR.

~~I'll commit new changes as soon as machinelib is ready~~

~~video: https://drive.google.com/file/d/1GRntQKIGSufXvWAOAYiXxdHcAwTsWR6r/view?usp=sharing~~ (outdated vid)


src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenDetectorBlockEntity.java
```diff
@@ -0,0 +1,160 @@
...
+
+    private void sendUpdate() {
+        setChanged(); // only for marking dirty, nothing to send.
+
+        if (level != null && !level.isClientSide) {
+            level.sendBlockUpdated(
+                worldPosition,
+                getBlockState(),
+                getBlockState(),
+                3
```
Roelymole: Please use the constant `Block.UPDATE_ALL` as it is more descriptive.

src/main/java/dev/galacticraft/mod/data/GCTranslationProvider.java
```diff
@@ -33,6 +34,32 @@
...
+import dev.galacticraft.mod.util.Translations.Misc;
+import dev.galacticraft.mod.util.Translations.RecipeCategory;
+import dev.galacticraft.mod.util.Translations.RegistryDebug;
+import dev.galacticraft.mod.util.Translations.SetOxygen;
+import dev.galacticraft.mod.util.Translations.SolarPanel;
+import dev.galacticraft.mod.util.Translations.SpaceRace;
+import dev.galacticraft.mod.util.Translations.Subtitles;
+import dev.galacticraft.mod.util.Translations.Tooltip;
+import dev.galacticraft.mod.util.Translations.Ui;
+import dev.galacticraft.mod.util.Translations.Waila;
```
Roelymole, line 62: Please remove all of these imports, as they are already being imported by the following line.

```java
import static dev.galacticraft.mod.util.Translations.*;
```

src/main/java/dev/galacticraft/mod/screen/GCMenuTypes.java
```diff
@@ -22,6 +22,8 @@
 
 package dev.galacticraft.mod.screen;
 
+import static dev.galacticraft.mod.content.GCBlocks.OXYGEN_DETECTOR;
```
Roelymole: Please remove this unused import.

src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenDetectorBlockEntity.java
```diff
@@ -0,0 +1,160 @@
...
+        CompoundTag tag = super.getUpdateTag(registryLookup);
+        tag.putBoolean("and", AND);
+
+        return tag;
+    }
+
+    @Override
+    protected @NotNull MachineStatus tick(@NotNull ServerLevel level, @NotNull BlockPos pos, @NotNull BlockState state,
+            @NotNull ProfilerFiller profiler) {
+        return null;
```
Roelymole, line 153: You shouldn't be returning `null` from a method marked `@NonNull`, instead you should return meaningful statuses such as `GCMachineStatuses.OXYGEN_DETECTED` and `GCMachineStatuses.NO_OXYGEN_DETECTED`.

I think moving the logic from the `isOxygenPresent()` method (without the `direction` argument) inside here would be a good starting point.

src/main/java/dev/galacticraft/mod/data/GCTranslationProvider.java
```diff
@@ -24,6 +24,7 @@
 
 import dev.galacticraft.api.data.TranslationProvider;
 import dev.galacticraft.api.rocket.part.RocketPart;
+import dev.galacticraft.mod.Constant.Cape;
```
Roelymole, line 27: Please remove this import, as it is already being imported by the following line.

```java
import static dev.galacticraft.mod.Constant.Cape;
```

src/main/java/dev/galacticraft/mod/client/gui/screen/ingame/OxygenDetectorScreen.java
```diff
@@ -0,0 +1,76 @@
...
+        int target_y = menu.be.isAnd() ? 0 : 34;
+        graphics.blit(Constant.ScreenTexture.OXYGEN_DETECTOR_SCREEN,
+            switchButton.getX(), switchButton.getY(),
+            176, target_y,
+            switchButton.getWidth(), switchButton.getHeight()
+        );
+
+        if (mouseIn(mouseX, mouseY, switchButton.getX(), switchButton.getY(), switchButton.getWidth(), switchButton.getHeight())) {
+            String mode_str = Component.translatable(menu.be.isAnd() ? Translations.Ui.DETECTOR_AND : Translations.Ui.DETECTOR_OR).getString();
+            Component mode = Component.translatable(Translations.Ui.DETECTOR_MODE, mode_str);
```
Roelymole: The `Component.translatable` method can accept other `Component` objects as arguments, so you don't need to convert to a string.

src/main/java/dev/galacticraft/mod/data/model/GCModelProvider.java
```diff
@@ -403,6 +403,7 @@ public void generateBlockStateModels(BlockModelGenerators generator) {
         );
 
         createFullCubeActiveMachine(generator, GCBlocks.OXYGEN_COLLECTOR);
+        createFullCubeActiveMachine(generator, GCBlocks.OXYGEN_DETECTOR);
```
Roelymole, line 406: The active model isn't currently being used, so if we don't need it, then it would be better to do something like this to generate the blockstate/model for the oxygen detector.

```java
MachineModelGenerator.createTrivialMachine(generator, GCBlocks.OXYGEN_DETECTOR, 
        TextureProvider.all(TextureMapping.getBlockTexture(GCBlocks.OXYGEN_DETECTOR))
);
```

However, perhaps we could put a little red dot at the bottom of the side textures (like the oxygen detector in Galacticraft 4) that gets lit up when the oxygen detector is outputting a redstone signal.

In which case, we would want to do something more like the following.

```java
createActiveMachine(generator, GCBlocks.OXYGEN_DETECTOR,
        TextureProvider.builder(Constant.MOD_ID)
                .sides("block/oxygen_detector_side_active")
                .top("block/oxygen_detector_top_active")
                .bottom("block/oxygen_detector_top_active")
                .build(),
        TextureProvider.builder(Constant.MOD_ID)
                .sides("block/oxygen_detector_side")
                .top("block/oxygen_detector_top")
                .bottom("block/oxygen_detector_top")
                .build()
);
```

src/main/java/dev/galacticraft/mod/content/item/GCCreativeModeTabs.java
```diff
@@ -291,6 +291,7 @@ public class GCCreativeModeTabs {
                 output.accept(REFINERY);
                 output.accept(FUEL_LOADER);
                 output.accept(OXYGEN_COLLECTOR);
+                output.accept(OXYGEN_DETECTOR);
```
Roelymole, line 294: The order that blocks are listed in REI is determined by the order in which they are registered in the `GCBlocks` class.

I think it is good to try to keep the order in the creative mode tabs consistent with this where possible.

My suggestion is to match the order in `GCBlocks`, or in other words, to put the oxygen detector between the oxygen compressor and the oxygen storage module.

Ideally the machines relating to oxygen would appear in a consistent order in most places, but I know this already isn't the case, so mainly just concern yourself with `GCBlockLootTableProvider`, `GCTranslationProvider` and `GCBlockTagsProvider`.

src/main/java/dev/galacticraft/impl/internal/mixin/oxygen/LevelChunkMixin.java
```diff
@@ -79,6 +84,22 @@ private LevelChunkMixin(ChunkPos pos, UpgradeData upgradeData, LevelHeightAccess
             }
             accessor.galacticraft$setInverted(x, y & 15, z, inverted);
         }
+
+        Boolean newValue = inverted ^ this.level.getDefaultBreathable();
```
Roelymole: You should use the primitive type `boolean` here instead of `Boolean`,  which is an object, to avoid a little bit of unnecessary overhead.

src/main/java/dev/galacticraft/impl/internal/mixin/oxygen/LevelChunkMixin.java
```diff
@@ -79,6 +84,22 @@ private LevelChunkMixin(ChunkPos pos, UpgradeData upgradeData, LevelHeightAccess
...
+
+            if (block instanceof OxygenChangeListener listener) {
+                listener.onOxygenChange(this.level, pos, direction.getOpposite(), newValue);
+            }
+
+            BlockEntity be = getBlockEntity(pos);
+            if (be instanceof OxygenChangeListener listener) {
+                listener.onOxygenChange(this.level, pos, direction.getOpposite(), newValue);
+            }
+        }
```
Roelymole, line 102: When a large volume of blocks is sealed or unsealed, then we may see the `onOxygenChange` method being invoked up to 6 times for each position, which is probably not the most efficient.

I think it would be better to check the surrounding positions as part of the `tick` method in `OxygenDetectorBlockEntity`, then this can be removed.

src/main/java/dev/galacticraft/mod/content/block/machine/OxygenDetectorBlock.java
```diff
@@ -0,0 +1,83 @@
...
+    protected boolean isSignalSource(BlockState state) {
+        return true;
+    }
+
+    @Override
+    protected int getSignal(BlockState state, BlockGetter world, BlockPos pos, Direction axisDirection) {
+        if (world.getBlockEntity(pos) instanceof OxygenDetectorBlockEntity entity) {
+            return entity.isOxygenPresent() ? 15 : 0;
+        }
+        return 0;
```
Roelymole, line 59: I have noticed that the redstone signal flickers when joining the world.

One possible solution could be to use a block state to keep track of whether the oxygen detector should output a signal even after reloading the world.

A bonus of changing the block state is that we can use this to update the neighbours without triggering chained updates.

The existing `ACTIVE` block state property could be used and this would not require calling `level.setBlock(...)` anywhere, as this is automatically updated based on the `MachineStatus` returned by the `tick` method of `OxygenDetectorBlockEntity` (there is a very minor fix in version 0.7.1+147 of MachineLib that you will need to get this to work).

Alternatively, you could add a flag to the `OxygenDetectorBlockEntity` that is persisted using the `saveAdditional` and `loadAdditional` methods, but this would require sending a separate update, so I don't think it would be so elegant.

src/main/java/dev/galacticraft/mod/util/Translations.java
```diff
@@ -368,6 +368,8 @@ interface SolarPanel {
     }
 
     interface MachineStatus {
+        String ENABLED = "ui.galacticraft.status.enabled";
+        String DISABLED = "ui.galacticraft.status.disabled";
```
Roelymole: Are these used anywhere?

src/main/java/dev/galacticraft/mod/screen/OxygenDetectorMenu.java
```diff
@@ -0,0 +1,47 @@
...
+    public OxygenDetectorMenu(int syncId, Inventory inv, BlockPos pos) {
+        super(GCMenuTypes.OXYGEN_DETECTOR, syncId, inv, pos, 8, 84);
+    }
+
+    @Override
+    public void cycleFaceConfig(BlockFace face, boolean reverse, boolean reset) {
+        super.cycleFaceConfig(face, reverse, reset);
+
+        this.be.sendUpdate(); // Redstone signal does not update when faces change in overworld. That's why I'm adding this.
+    }
```
Roelymole, line 46: If the `OxygenDetectorBlockEntity` checks for changes each tick, then this override might become unnecessary.

src/main/java/dev/galacticraft/mod/screen/GCMenuTypes.java
```diff
@@ -22,6 +22,8 @@
...
+
```
Roelymole, line 26: Please remove this unused import.

src/main/java/dev/galacticraft/mod/client/gui/screen/ingame/OxygenDetectorScreen.java
```diff
@@ -0,0 +1,88 @@
...
+    @Override
+    protected void renderMachineBackground(GuiGraphics graphics, int mouseX, int mouseY, float delta) {
+        int target_y = menu.be.isAnd() ? 0 : 34;
+        graphics.blit(Constant.ScreenTexture.OXYGEN_DETECTOR_SCREEN,
+            switchButton.getX(), switchButton.getY(),
+            176, target_y,
+            switchButton.getWidth(), switchButton.getHeight()
+        );
+
+        String mode_str = Component.translatable(menu.be.isAnd() ? Translations.Ui.DETECTOR_AND : Translations.Ui.DETECTOR_OR).getString();
```
Roelymole: The conversion to string here is unnecessary as `Component.translatable` accepts other `Component` objects as arguments.

src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenDetectorBlockEntity.java
```diff
@@ -0,0 +1,161 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+package dev.galacticraft.mod.content.block.entity.machine;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
```
Roelymole: It is the convention throughout the codebase to put these imports immediately after all of the `net.minecraft` imports (with no blank line) so that they are arranged alphabetically.

src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenDetectorBlockEntity.java
```diff
@@ -0,0 +1,161 @@
...
+        IOFace option = getIOConfig().get(BlockFace.from(this.getBlockState(), direction));
+        return option.getType() != ResourceType.OVERRIDE; // Override disables detecting oxygen. (and switches to base texture)
+    }
+
+    public boolean isOxygenPresent(Direction direction) {
+        BlockPos pos = this.worldPosition.relative(direction);
+        boolean isFree = !this.level.getBlockState(pos).isSolidRender(this.level, pos);
+        if(oxygenWorld) return isFree;
+
+        return this.level.isBreathable(pos) && isFree;
```
Roelymole, line 112: This can be rearranged slightly, making it a bit clearer that the return value is always `false` when `isFree` is `false`.

```suggestion
        return isFree && (oxygenWorld || this.level.isBreathable(pos));
```

src/main/java/dev/galacticraft/mod/client/gui/screen/ingame/OxygenDetectorScreen.java
```diff
@@ -0,0 +1,88 @@
...
+        Component mode = Component.translatable(Translations.Ui.DETECTOR_MODE, mode_str).withColor(getModeColor());
```
Roelymole: The `Component.translatable` method can accept other `Component` objects as arguments, so you don't need to convert to a string.

src/main/java/dev/galacticraft/mod/data/GCTranslationProvider.java
```diff
@@ -1137,6 +1166,8 @@ protected void generateSolarPanelTranslations() {
     }
 
     protected void generateMachineStatusTranslations() {
+        this.add(MachineStatus.ENABLED, "Enabled");
+        this.add(MachineStatus.DISABLED, "Disabled");
```
Roelymole, line 1170: These translations aren't used anywhere.

src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenDetectorBlockEntity.java
```diff
@@ -0,0 +1,161 @@
...
+
+    @Override
+    public boolean faceHasOverride(BlockFace face) {
+        return true;
+    }
+
+    public boolean isFaceActive(Direction direction) {
+        IOFace option = getIOConfig().get(BlockFace.from(this.getBlockState(), direction));
+        return option.getType() != ResourceType.OVERRIDE; // Override disables detecting oxygen. (and switches to base texture)
+    }
```
Roelymole: It might be worth inverting this method to avoid inverting it back again in the only place that it is currently being used.

```suggestion
    public boolean isFaceDisabled(Direction direction) {
        IOFace option = getIOConfig().get(BlockFace.from(this.getBlockState(), direction));
        return option.getType() == ResourceType.OVERRIDE; // Override disables detecting oxygen. (and switches to base texture)
    }
```

src/main/java/dev/galacticraft/mod/content/block/machine/OxygenDetectorBlock.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Override
+    public void neighborChanged(BlockState state, Level level, BlockPos pos, Block block, BlockPos fromPos,
+            boolean notify) {
+        super.neighborChanged(state, level, pos, block, fromPos, notify);
+        level.updateNeighborsAt(pos, this);
+    }
```
Roelymole, line 67: Placing two oxygen detectors next to each other will prevent block interactions from working for a while with the following error message.

```
[22:32:59] [Server thread/ERROR] (Minecraft) Too many chained neighbor updates. Skipping the rest. First skipped position: -246, 64, -249
```

The override of this method looks like the cause of the problem.

src/main/java/dev/galacticraft/mod/client/gui/screen/ingame/OxygenDetectorScreen.java
```diff
@@ -0,0 +1,88 @@
...
+        graphics.drawString(font, mode, this.leftPos + 12, this.topPos + 24, mode.getStyle().getColor().getValue());
+
+        Component stat = Component.translatable(menu.be.isOxygenPresent() ? Translations.Ui.OXYGEN_DETECTED : Translations.Ui.NO_OXYGEN_DETECTED);
```
Roelymole, line 81: Recalculating whether there is oxygen present is not ideal and may get out of sync between the client and the server.

It would be better to perform this calculation once in the `tick` method of `OxygenDetectorBlockEntity` and return a meaningful `MachineStatus` corresponding to the result (which should be defined in `GCMachineStatuses`).

The `MachineStatus` objects contain a colour that is automatically applied to the `Component`, so the `getStatColor` method may be removed as well.

The following example shows how this is then accessed.

https://github.com/TeamGalacticraft/Galacticraft/blob/1f0ab90312ab84d66d2efe4e0e18407bd0c2d6ff/src/main/java/dev/galacticraft/mod/client/gui/screen/ingame/OxygenCollectorScreen.java#L54-L55

src/main/java/dev/galacticraft/mod/client/gui/screen/ingame/OxygenDetectorScreen.java
```diff
@@ -0,0 +1,88 @@
...
+        int target_y = menu.be.isAnd() ? 0 : 34;
+        graphics.blit(Constant.ScreenTexture.OXYGEN_DETECTOR_SCREEN,
+            switchButton.getX(), switchButton.getY(),
+            176, target_y,
+            switchButton.getWidth(), switchButton.getHeight()
+        );
+
+        String mode_str = Component.translatable(menu.be.isAnd() ? Translations.Ui.DETECTOR_AND : Translations.Ui.DETECTOR_OR).getString();
+        Component mode = Component.translatable(Translations.Ui.DETECTOR_MODE, mode_str).withColor(getModeColor());
+        graphics.drawString(font, mode, this.leftPos + 12, this.topPos + 24, mode.getStyle().getColor().getValue());
```
Roelymole: Other machines do not use a shadow and I personally find it harder to read, so it would be good to be consistent and turn it off by providing an extra argument to the `drawString` method like follows.

```suggestion
        graphics.drawString(font, mode, this.leftPos + 12, this.topPos + 24, mode.getStyle().getColor().getValue(), false);
```

src/main/java/dev/galacticraft/mod/client/gui/screen/ingame/OxygenDetectorScreen.java
Roelymole, line 1: Personally, I find this text quite hard to read with the choice of colours in combination with the shadow.

Also, the other machines do not use a shadow, so it would be good to be consistent and turn it off.

I said it before, but I really think we should consider alternatives to this switch texture as it doesn't really fit.

src/main/java/dev/galacticraft/mod/client/gui/screen/ingame/OxygenDetectorScreen.java
Roelymole, line 79: The `Component.translatable` method can accept other `Component` objects as arguments, so you don't need to convert to the string `mode_str`.

Also, I don't think the value of `mode.getStyle().getColor().getValue()` is actually used, since the component already has a colour, so it might be neater if this was replaced with -1 (which is something that is done extensively in the vanilla code).

The shadow can be turned off by using one of the overloads of the `drawString` method by providing an extra argument with the value of `false`.

e.g. `drawString(Font font, Component component, int x, int y, int color, boolean shadow)`

```suggestion
        Component mode = Component.translatable(menu.be.isAnd() ? Translations.Ui.DETECTOR_AND : Translations.Ui.DETECTOR_OR);
        mode = Component.translatable(Translations.Ui.DETECTOR_MODE, mode).withColor(getModeColor());
        graphics.drawString(font, mode, this.leftPos + 12, this.topPos + 24, -1, false);
```

src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenDetectorBlockEntity.java
Roelymole, line 1: The following overrride should be added to ignore the settings in the redstone configuration tab as choosing the middle option may cause the oxygen detector to rapidly turn itself on and off when oxygen is detected.

I might look more into what to do with this tab on the MachineLib side.

```java
@Override
public boolean isDisabled() {
    // Ignore the settings in the redstone configuration tab
    return false;
}
```

src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenDetectorBlockEntity.java
```diff
@@ -0,0 +1,161 @@
...
+
+    public boolean isOxygenPresent(Direction direction) {
```
Roelymole: It might be good to make this into a private helper method.

src/main/java/dev/galacticraft/mod/client/gui/screen/ingame/OxygenDetectorScreen.java
```diff
@@ -0,0 +1,88 @@
...
+        switchButton = new Button(this.leftPos + 121, this.topPos + 24, 24, 33, Component.empty(), btn -> {invertMode();}, s -> s.get()) {
+            @Override
+            protected void renderWidget(GuiGraphics graphics, int mouseX, int mouseY, float delta) {};
+        };
+        addRenderableWidget(switchButton);
+    }
+
+    private int getModeColor() {
+        return menu.be.isAnd() ? 0xFF0000 : 0x00FF00;
+    }
```
Roelymole, line 62: I think it would be better to choose from the standard set colours in `ChatFormatting`, which would be more descriptive than just a hex code.

There is no need to convert to an integer as `.withColor(getModeColor())` can just be replaced with `.withStyle(getModeColor())`

src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenDetectorBlockEntity.java
```diff
@@ -0,0 +1,161 @@
...
+import net.minecraft.nbt.CompoundTag;
+import net.minecraft.server.level.ServerLevel;
+import net.minecraft.util.profiling.ProfilerFiller;
+import net.minecraft.world.entity.player.Inventory;
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.level.Level;
+import net.minecraft.world.level.block.state.BlockState;
+
+public class OxygenDetectorBlockEntity extends MachineBlockEntity {
+    private boolean AND = false;
```
Roelymole, line 50: It is conventional to use all uppercase for constant variables, so calling it something like `andMode` might be better.

src/main/java/dev/galacticraft/mod/network/c2s/OxygenDetectorControlPayload.java
```diff
@@ -0,0 +1,64 @@
...
+import dev.galacticraft.machinelib.impl.Constant;
+import dev.galacticraft.mod.screen.OxygenDetectorMenu;
+import io.netty.buffer.ByteBuf;
+import net.fabricmc.fabric.api.networking.v1.ServerPlayNetworking;
+import net.minecraft.network.codec.StreamCodec;
+import net.minecraft.network.protocol.common.custom.CustomPacketPayload;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.server.level.ServerPlayer;
+
+public record OxygenDetectorControlPayload(boolean mode) implements C2SPayload {
```
Roelymole, line 37: It would suggest renaming this argument to `andMode` so that it is clearer what the value of `true` represents.

src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenDetectorBlockEntity.java
```diff
@@ -0,0 +1,161 @@
...
+    public boolean isAnd() {
+        return AND;
+    }
+
+    @Override
+    public boolean faceHasOverride(BlockFace face) {
+        return true;
+    }
+
+    public boolean isFaceActive(Direction direction) {
```
Roelymole, line 102: It might be good to make this into a private helper method.

src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenDetectorBlockEntity.java
Roelymole, line 107: This could also probably be made into a private helper method.

src/main/java/dev/galacticraft/mod/content/block/entity/machine/OxygenDetectorBlockEntity.java
```diff
@@ -0,0 +1,161 @@
...
+
+        if (level != null && !level.isClientSide) {
+            level.sendBlockUpdated(
+                worldPosition,
+                getBlockState(),
+                getBlockState(),
+                3
+            );
+            level.updateNeighborsAt(worldPosition, getBlockState().getBlock());
+        }
```
Roelymole, line 81: This might become unnecessary.

Roelymole (changes requested):
This seems to be working for the most part, thank you. However, there are some simplifications that can be made.

An overarching theme of these comments is that you should be using the `tick` method of `OxygenDetectorBlockEntity` to test whether the surrounding positions have oxygen, which will allow MachineLib to handle lots of the work.

Hopefully I haven't been too confusing, but if there is anything you aren't clear about, please ask and I will try to clarify.

PR #594, opened 2026-09-13
fix: prevent blocks being extinguished in sealed rooms and add gametests

Roelymole:
Added three gametests:
- Test that a torch is extinguished when it is placed in a position without oxygen.
- Test that a player can reignite an unlit torch using a flint and steel.
- Test that a torch does not get extinguished when in a sealed room (expecting this to fail to begin with).


TakeABao23 (approved):
It looks great! No notes.

Roelymole:
I have been thinking of increasing the gametest coverage to include the other extinguishable blocks.

The current tests with the torches are very quick, so it might not be a problem having more of them.

I haven't quite thought of the nearest way to do this yet without repeating too much code.

TakeABao23:
Would there be a way to reuse the setup of the torch test and just replace the torch with the extinguishable block?

Roelymole:
It is definitely possible, but I haven't settled on something that I am quite happy with.

TakeABao23:
How hard would it be to add jack o lantern extinguishing to this while you're at it? Or would that be big enough to be a different PR?

Roelymole:
I've had an idea for making the Jack O'Lanterns extinguishable, which roughly speaking would be to make them permeable to oxygen (possibly only in one direction, depending on how convenient that is).

It should be a relatively small task, but it doesn't have to be part of this PR.

The current testing doesn't cover all of the extinguishable blocks yet either, as the coal powered machines and fire blocks behave a bit differently.

